PR #95, opened 2026-03-10
Improve MCP tool selection priority for AI agents

dpage:
AI coding agents (Claude Code, Cursor, etc.) tend to prefer shell commands like psql over MCP tools because their training data contains more examples of shell-based database workflows. This makes users manually tell agents to use MCP tools, which is a poor experience.

This commit addresses the problem through three complementary changes:

1. Server-level instructions: Add an `instructions` field to the MCP initialize response per the MCP spec. This is the strongest lever available to MCP server authors — the instructions are always present in the client context without the "ignore if irrelevant" caveat that applies to CLAUDE.md content.

2. Assertive tool descriptions: Update all tool descriptions to include explicit "use this instead of X" language. Models read these when deciding which tool to call, and the direct comparison with competing alternatives steers selection toward the MCP tools.

3. Agent integration documentation: Add a new docs page covering server-level instructions, recommended CLAUDE.md/cursorrules configuration, and tips for maintaining consistent tool selection.

## Summary by CodeRabbit

* **New Features**
  * Server now sends initialization instructions to guide AI agents toward using MCP tools instead of psql/shell commands.

* **Documentation**
  * Added a guide for configuring AI agent tool selection with example project-level files and tips.
  * Updated MCP tool descriptions to recommend using the corresponding tools (e.g., counting, schema, explain, embeddings, search, DB switching) instead of manual psql/shell usage.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,14 @@ and this project adheres to
 
 ### Added
 
+- MCP tool selection guidance for AI agents. The server now sends
+  a server-level `instructions` field during the MCP initialize
+  handshake, directing agents to prefer MCP tools over `psql` and
+  shell commands. Tool descriptions include explicit "use this
+  instead of..." language to steer tool selection. A new
+  documentation page covers recommended `CLAUDE.md` and
+  `.cursorrules` configuration for reinforcing tool preference.
+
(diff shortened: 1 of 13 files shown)


rshoemaker (approved):
Looks good.

PR #98, opened 2026-03-12
Container deployment related fixes to address #93.

### LINKED ISSUE #93: Feedback on the project
I have built a project based on this mcp server. Check it here: https://github.com/Milad/Open-Memory

Thank you very much for your work. I'd like to provide my feedback:
- Would have loved to use an already-dockerized image in my composer file, instead of having to clone the branch and build from it.
- It's not easy and directly clear how to enable write access to the DB. In a project like mine, writing to the DB is essential. Maybe, it's better to allow write access for non-numbered DB configurations?
- Secrets are printed. For example, I saw the database password and the auth token secret in the output of the mcp server. Would be more secure to hide them.

dpage:
fixes #93

## Summary by CodeRabbit

* **New Features**
  * Control database write operations (INSERT/UPDATE/DELETE) via a new PGEDGE_DB_ALLOW_WRITES environment variable (disabled by default)
  * Pre-built container images available on GitHub Container Registry for simplified Docker deployments

* **Documentation**
  * Updated Docker deployment guide to reference pre-built images and improve steps
  * Expanded configuration docs to cover the new write-access environment variable

### DIFF
--- a/docker/init-server.sh
+++ b/docker/init-server.sh
@@ -82,8 +82,10 @@ YAML_DATABASES
 
         # Convert allow_writes to boolean
         local db_allow_writes
-        case "$PGEDGE_DB_ALLOW_WRITES" in
-            true|TRUE|1|yes|YES) db_allow_writes="true" ;;
+        local allow_writes_lower
+        allow_writes_lower=$(echo "$PGEDGE_DB_ALLOW_WRITES" | tr '[:upper:]' '[:lower:]')
+        case "$allow_writes_lower" in
+            true|1|yes) db_allow_writes="true" ;;
             *) db_allow_writes="false" ;;
         esac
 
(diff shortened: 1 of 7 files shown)


docker/init-server.sh
```diff
@@ -79,6 +79,14 @@ YAML_DATABASES
     if [ -n "$PGEDGE_DB_HOST" ] || [ -n "$PGEDGE_DB_NAME" ]; then
         DB_COUNT=$((DB_COUNT + 1))
         local db_name="${PGEDGE_DB_NAME:-default}"
+
+        # Convert allow_writes to boolean
+        local db_allow_writes
+        case "$PGEDGE_DB_ALLOW_WRITES" in
+            true|TRUE|1|yes|YES) db_allow_writes="true" ;;
```
rshoemaker: Looks like there's a mismatch in behavior between the init script and the go code.  You accept "true|TRUE|1|yes|YES" here, but only "true|1|yes" in internal/config/config.go:665.  It would be safer to normalize the input in both places toLower or at least keep the behavior exactly the same.

Using PGEDGE_DB_ALLOW_WRITES=TRUE would work in Docker, but not when running the go binary directly.

rshoemaker (approved):
Looks good, just the one question below about normalization.

PR #100, opened 2026-03-13
Fix workflow_dispatch builds by using GoReleaser snapshot mode

maqeel75:
Skip GitHub release creation and use --snapshot flag when triggered
  manually, so builds succeed without a tagged commit.

## Summary by CodeRabbit

* **Chores**
  * Enable on-demand snapshot builds when the release workflow is manually triggered, producing snapshot artifacts for testing.
  * Ensure the main release job runs only for tagged pushes, so official releases are created from tags.
  * Allow the release creation step to execute for manual triggers and pushes, so release artifacts and notes can be generated outside tag-only flows.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,7 +128,7 @@ jobs:
               with:
                   distribution: goreleaser
                   version: latest
-                  args: release --clean --config .goreleaser-amd64.yaml ${{ github.event_name == 'workflow_dispatch' && '--snapshot' || '' }}
+                  args: release --clean --config .goreleaser-amd64.yaml ${{ github.event_name == 'workflow_dispatch' && github.ref_type != 'tag' && '--snapshot' || '' }}
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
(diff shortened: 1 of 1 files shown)


dpage (commented):
## Review

The change to add `--snapshot` to the GoReleaser args on `workflow_dispatch` is correct for the build jobs. However, the downstream `release` job (lines 234-306) still runs unconditionally and will fail or misbehave during manual runs:

### 🔴 Must Fix: `release` job needs to be skipped for `workflow_dispatch`

During a `workflow_dispatch` run:

1. **Line 274**: `${GITHUB_REF_NAME#v}` will use a branch name instead of a version, producing archives like `pgedge-nla-web_main_noarch.tar.gz`
2. **Line 282**: `sha256sum *.zip` may fail if snapshot mode doesn't produce `.zip` files
3. **Lines 285-296**: `softprops/action-gh-release` will attempt to create a GitHub release without a tag, which will either fail or create an unwanted release

**Suggested fix** — add a condition to skip the `release` job on `workflow_dispatch`:

```yaml
release:
    name: Create Release
    runs-on: ubuntu-latest
    needs: [build-amd64, build-arm64]
    if: github.event_name != 'workflow_dispatch'
```

### 🟡 Should Fix: Verify snapshot artifact paths

GoReleaser snapshot mode may change output paths (e.g., adding `-SNAPSHOT-` suffixes to artifact names). The upload steps at lines 142-144 and 228-230 use `dist/*.tar.gz` and `dist/*.zip` globs which should still match, but worth verifying with a test run.

dpage (approved):

PR #112, opened 2026-03-26
Strip semicolons from queries before appending LIMIT/OFFSET. #110

dpage:
#110

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where SQL queries ending with trailing semicolons could cause syntax errors when LIMIT/OFFSET were appended; such queries are now normalized so appended clauses work correctly.

* **Documentation**
  * Updated the changelog to note the trailing-semicolon handling improvement.

* **Tests**
  * Added unit tests covering many trailing-semicolon scenarios to ensure correct normalization and preservation of semicolons inside string literals.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,10 @@ and this project adheres to
 
 ### Fixed
 
+- Queries with trailing semicolons no longer produce a SQL syntax
+  error when the server auto-appends a `LIMIT` clause. The server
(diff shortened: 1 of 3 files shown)


rshoemaker (approved):
Coderabbit called out an issue with interleaved spaces and semis not being handled correctly by your diff, so I pushed a commit that fixes that and adds some tests.

Take a look at my commit before you merge to make sure it looks ok to you.

PR #143, opened 2026-05-04
fix(http): return 202 Accepted with empty body for JSON-RPC notifications

### LINKED ISSUE #142: HTTP transport: server replies to JSON-RPC notifications with malformed (no-`id`) Response objects
## Summary

When the server receives a JSON-RPC **notification** (a request with no `id`) on `/mcp/v1`, it replies with `HTTP 200 OK` and a JSON-RPC Response/Error body that has no `id` field. Per JSON-RPC 2.0 §4.1, the server must not reply to notifications at all; per the MCP streamable HTTP spec, it must return `202 Accepted` with an empty body. The reply is also itself malformed (a Response/Error object with no `id` is not a valid JSON-RPC message), which causes strict clients to throw on every notification.

## Version

- `ghcr.io/pgedge/postgres-mcp:1.0.0`
- HTTP transport on `/mcp/v1`

## Repro

```bash
# Notification: notifications/initialized (the post-initialize notification every MCP client sends)
curl -i -X POST http://localhost:8080/mcp/v1 \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  --data '{"jsonrpc":"2.0","method":"notifications/initialized","params":{}}'
```

Server response:

```http
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 30

{"jsonrpc":"2.0","result":{}}
```

Same shape with `notifications/cancelled` and `notifications/roots/list_changed`, except the body is an error envelope instead of a result envelope:

```http
HTTP/1.1 200 OK
Content-Type: application/json

{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"}}
```

Both bodies are missing `id` (and have no `method` either, so they aren't requests/notifications). They don't match any valid JSON-RPC 2.0 message shape.

For comparison, a regular request with `id` works correctly:

```bash
curl -i -X POST http://localhost:8080/mcp/v1 \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  --data '{"jsonrpc":"2.0","id":42,"method":"tools/list","params":{}}'
```

```http
HTTP/1.1 200 OK
Content-Type: application/json

{"jsonrpc":"2.0","id":42,"result":{"tools":[…]}}
```

## Expected

Per the MCP streamable HTTP spec ("[Sending Messages to the Server](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server)"):

> If the input consists solely of (any number of) JSON-RPC responses or notifications:
> - If the server accepts the input, the server **MUST** return HTTP status code 202 Accepted with no body.

And per [JSON-RPC 2.0 §4.1](https://www.jsonrpc.org/specification#notification):

> A Notification is a Request object without an "id" member. … The Server **MUST NOT** reply to a Notification, including those that are within a batch request.

Expected response to any of the notification posts above:

```http
HTTP/1.1 202 Accepted
Content-Length: 0
```

## Impact on clients

Strict JSON-RPC 2.0 deserializers reject the malformed bodies. Concretely, the .NET MCP SDK (`ModelContextProtocol.Client.StreamableHttpClientSessionTransport` + `ModelContextProtocol.Protocol.JsonRpcMessage.Converter`) is a polymorphic JSON converter that picks the message subtype by shape: presence of `method` → Request or Notification, presence of `result` or `error` → Response or ErrorResponse (which must have `id`). A body with `result` (or `error`) but no `id` matches no valid subtype, so the converter throws:

```
System.Text.Json.JsonException: Invalid JSON-RPC message format
   at ModelContextProtocol.Protocol.JsonRpcMessage.Converter.Read(...)
   at ModelContextProtocol.Client.StreamableHttpClientSessionTransport.ProcessMessageAsync(...)
```

The client recovers (the underlying tool calls still work since regular `id`-bearing responses parse fine), but every session emits this on the post-`initialize` `notifications/initialized` and intermittently on other notifications. Noisy in logs, and a strict client could legitimately drop the connection on receiving a malformed message.

## Suggested fix

In the streamable HTTP handler, branch on whether the parsed inbound message has an `id`:
- `id is None` → it's a notification: return `202 Accepted` with empty body, do not invoke the response builder.
- `id is not None` → existing behavior.

Same branch should be applied to unknown method handling — sending `-32601 Method not found` for a notification is doubly wrong (replying when forbidden, and replying without `id`).

## Side observation (not part of this bug)

`GET /mcp/v1` returns `405 Method Not Allowed`. This is conformant — the spec explicitly allows `405` for servers that don't offer server-initiated SSE pushes. No action needed; just noting it for completeness.

## References

- JSON-RPC 2.0 §4.1: https://www.jsonrpc.org/specification#notification
- MCP streamable HTTP transport: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server

richyen:
Per JSON-RPC 2.0 §4.1, the server MUST NOT reply to a Notification (a Request object without an "id" member). Per the MCP streamable HTTP transport spec, the server MUST return 202 Accepted with no body for notifications. The previous behavior returned a 200 with a JSON-RPC Response/Error body that had no "id" field — which is itself not a valid JSON-RPC message and causes strict clients (e.g. the .NET MCP SDK polymorphic JsonRpcMessage converter) to throw on every notification.

Changes:
- handleHTTPRequest now inspects the raw request body for the presence of an "id" member and short-circuits notifications before dispatch, responding with 202 and an empty body. This applies uniformly to known notification methods (notifications/initialized) and unknown ones, closing the secondary bug where unknown notifications received a -32601 "Method not found" reply.
- Introduced hasIDField helper to distinguish "id absent" (notification, no reply) from "id": null (a request, must reply) — Go's interface{} unmarshaling collapses both to nil.
- Removed the now-dead notifications/initialized case in the dispatch switch; notifications no longer reach handleRequestHTTP.
- Updated TestHandleNotificationsInitialized to assert 202 + empty body for a real notification (the previous version sent an id, exercising the buggy path it claimed to verify).
- Added TestHandleNotification_UnknownMethod (covers /cancelled, /roots/list_changed, and any unknown notification) and TestHandleNotification_NullIDIsRequest (regression guard ensuring "id": null is treated as a request, not a notification).

Fixes #142

## Summary by CodeRabbit

* **Bug Fixes**
  * JSON-RPC notifications now return HTTP 202 Accepted with an empty body.
  * Better distinction between notifications (no id) and requests with explicit id: null.
  * Unknown notification methods are acknowledged silently (no error reply).

* **New Features**
  * Schema metadata cache now auto-refreshes with a configurable TTL (default: 5 minutes).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,15 @@ and this project adheres to
 
 ### Fixed
 
+- HTTP transport now returns `202 Accepted` with an empty body for
+  JSON-RPC notifications, per JSON-RPC 2.0 §4.1 and the MCP streamable
+  HTTP transport spec. Previously, the server replied to notifications
+  with a `200 OK` response that had no `id` field, which is itself not
+  a valid JSON-RPC message and caused strict clients (such as the .NET
+  MCP SDK) to throw on every notification. Unknown notification methods
+  are now also acknowledged silently rather than receiving a `-32601`
+  error reply. (#142)
+
 - Database switching via `select_database_connection` now persists
   correctly in HTTP mode for unbound API tokens.
   `GetAccessibleDatabases` previously returned only the first
--- a/internal/mcp/http_server.go
+++ b/internal/mcp/http_server.go
@@ -189,6 +189,12 @@ func (s *Server) handleHTTPRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Detect a JSON-RPC notification: a Request object without an "id" member
+	// (per JSON-RPC 2.0 §4.1). Note that "id": null is a valid request id and
+	// is NOT a notification, so we must inspect the raw JSON — interface{}
+	// unmarshaling collapses "absent" and "null" to the same nil value.
+	isNotification := !hasIDField(body)
+
 	// Set up tracing context
 	tokenHash := auth.GetTokenHashFromContext(ctx)
 	sessionID := tokenHash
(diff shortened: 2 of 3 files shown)


dpage (approved):

dpage:
Thank you for the PR @richyen!

PR #144, opened 2026-05-04
fix(database): tolerate NULL columns in metadata LEFT JOIN

### LINKED ISSUE #126: Metadata loader fails with NULL column_name when database contains a table with no columns
Description:
The MCP server throws no database connection configured for this token: failed to load metadata for database 'default': failed to scan row: can't scan into dest[6] (col: column_name): cannot scan NULL into *string when the connected database contains a table with no columns.

Steps to Reproduce:

Create a table with no columns in PostgreSQL:

- CREATE TABLE empty_columns_test();
- Start the MCP server pointing to that database
- Make any query that triggers metadata loading

docker logs postgres-mcp --tail 20
  
  
```
Status:   Enabled
======================================================================

Created users file with 1 user(s)
Starting MCP server
Loaded 0 API token(s) from /app/data/tokens.json
Watching /app/data/tokens.json for changes
Loaded 1 user(s) from /app/data/users.json
Watching /app/data/users.json for changes
Rate limiting enabled: 10 attempts per 15 minutes per IP
Database configured: postgres@host.docker.internal:6000/postgres (per-session connections)
Authentication: ENABLED
Conversation store: /app/data/conversations.db
Starting MCP server in HTTP mode on :8080
Authentication: ENABLED
LLM Proxy: ENABLED (provider: anthropic, model: claude-sonnet-4-20250514)
Knowledgebase: DISABLED
Conversation history: ENABLED
2026/04/14 17:10:31 [AUTH] Reloaded /app/data/users.json
ERROR: Failed to get database client for tool 'get_schema_info': no database connection configured for this token: failed to load metadata for database 'default': failed to scan row: can't scan into dest[6] (col: column_name): cannot scan NULL into *string
```

Root Cause:
In internal/database/connection.go:484, columnName, dataType, and isNullable are scanned into plain string variables. When a table has no columns, the LEFT JOIN returns NULL for these fields, which pgx cannot scan into a non-pointer string.

Expected Behavior:
Tables with no columns should be skipped gracefully without failing the entire metadata load.

Actual Behavior:
The entire database connection fails with a NULL scan error, surfaced to the user as a misleading no database connection configured error.

Fix:
Change columnName, dataType, and isNullable to *string and skip rows where columnName is nil.

richyen:
The metadata loader's main query LEFT JOINs the per-table catalog rows against column_info. When the database contains a table with zero columns (e.g. CREATE TABLE empty_columns_test()), the LEFT JOIN produces a row whose ci.column_name, ci.data_type, and ci.is_nullable are all NULL. The row scan declared those targets as plain string, so pgx aborts with:

  failed to scan row: can't scan into dest[6] (col: column_name):
  cannot scan NULL into *string

That single bad row failed the entire metadata load, which surfaced to callers as the misleading 'no database connection configured for this token' error and blocked every tool that depends on metadata (get_schema_info, similarity_search, etc.).

Fix: scan columnName/dataType/isNullable as sql.NullString (matching how typeName is already handled in the same scan) and gate the existing skip-row check on columnName.Valid. Tables with zero columns now appear in the metadata with an empty Columns slice and no longer poison the load.

Adds TestLoadMetadata_TableWithNoColumns, an integration-style regression test that creates an empty-columns table against the database referenced by TEST_PGEDGE_POSTGRES_CONNECTION_STRING (the same env var used elsewhere in this repo) and asserts the load succeeds and the table is present with zero columns. Skipped when the env var is unset.

Fixes #126

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,17 @@ and this project adheres to
 
 ### Fixed
 
+- Metadata loader now tolerates tables with zero columns
+  (e.g. `CREATE TABLE foo()`). The query LEFT JOINs against the
+  per-column catalog, so a zero-column table produced a row whose
+  `column_name`, `data_type`, and `is_nullable` were all NULL; the
+  scan declared those targets as plain `string` and aborted with
+  `cannot scan NULL into *string`, failing the entire metadata load
+  and surfacing as the misleading `no database connection
+  configured for this token` error. The three columns are now
+  scanned as `sql.NullString` and zero-column tables appear in the
+  metadata with an empty `Columns` slice. (#126)
+
 - Database switching via `select_database_connection` now persists
   correctly in HTTP mode for unbound API tokens.
   `GetAccessibleDatabases` previously returned only the first
--- a/internal/database/connection.go
+++ b/internal/database/connection.go
@@ -482,7 +482,13 @@ func (c *Client) LoadMetadataFor(connStr string) error {
 	columnCount := 0
 
 	for rows.Next() {
-		var schemaName, tableName, tableType, tableDesc, columnName, dataType, isNullable, columnDesc string
+		// columnName, dataType, and isNullable come from the LEFT JOIN against
+		// column_info; they are NULL for tables that have zero columns
+		// (e.g. CREATE TABLE foo();). Scan them as NullString so the row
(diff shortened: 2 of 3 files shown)


dpage (approved):

dpage:
Thanks for the PR!

PR #168, opened 2026-05-27
fix: machine-safe built-in resource name + env vars for builtins (#139)

### LINKED ISSUE #139: Interoperability issue with built-in resource names when used through MCP clients
Hi!

I think I may have run into an interoperability issue with pgedge-postgres-mcp when it’s used through another MCP client layer that forwards MCP capabilities as tool definitions to Anthropic.

I’m running pgedge-postgres-mcp v1.0.0 over HTTP, and Nanobot is connecting to it successfully. In the Nanobot logs I can see the pgEdge MCP server connecting and registering tools, resources, and prompts correctly. For example, it registers tools like query_database, get_schema_info, etc., and also a resource that ends up showing up as:
```
mcp_postgres_mcp_resource_PostgreSQL System Information
```

So the MCP connection itself appears healthy.

The problem only shows up when I send a request through Nanobot using Claude Sonnet. Anthropic rejects the tool list with an error like:
```
tools.21.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'
```

My suspicion is that the built-in resource name “PostgreSQL System Information” is the trigger here, because it contains spaces, and a downstream client is apparently turning that into a provider tool name without sanitizing it first.

To be clear, I realize the downstream client may share responsibility here; it probably should sanitize names before forwarding them. But I wanted to raise this here too, because the built-in resource/prompt names seem to be optimized for readability rather than for stricter provider tool-name rules, and that can break interoperability in multi-hop MCP setups.

A couple things that might help:
- expose a machine-safe identifier for built-in resources/prompts in addition to the human-readable label
- avoid spaces/special characters in exported built-in capability names
- allow built-in resources/prompts to be disabled via environment variables as well as config file

That last point would have helped as a workaround for containerized deployments. From what I could tell in `internal/config/config.go`, built-ins are only configurable via config file, not env vars.

Happy to provide logs/config details if useful. I mainly wanted to report that the server works, but one of the built-in capability names seems to cause trouble with Anthropic-based tool validation downstream.

I will try OpenAI models as well.

dpage:
## Summary

- Rename the built-in `pg://system_info` resource from
  `"PostgreSQL System Information"` to `postgresql_system_info` so the
  name matches the identifier pattern enforced by Anthropic's
  tool-name validation (`^[a-zA-Z0-9_-]{1,128}$`). Downstream MCP
  clients that forward built-in capability names as provider tool
  names (e.g. Nanobot -> Claude) were rejecting the original name
  because of the embedded spaces. The resource URI is unchanged.
- Add `PGEDGE_BUILTIN_TOOL_*`, `PGEDGE_BUILTIN_RESOURCE_*`, and
  `PGEDGE_BUILTIN_PROMPT_*` environment variables for toggling
  individual built-ins, which the issue called out as useful for
  containerized deployments where editing the config file is awkward.
- Update docs (`configuration.md`, `feature_config.md`,
  `mcp-protocol.md`, `changelog.md`) and tests.

## Test plan

- [x] `go test ./...` (unit tests pass — integration tests under
  `./test/` require a working postgres and were already failing
  locally without one)
- [x] `make lint-server` (0 issues)
- [x] `gofmt -l .` (clean)
- [x] New unit tests cover `setBoolPtrFromEnv` and verify
  `applyEnvironmentVariables` wires every builtin env var
- [x] Integration server output shows the renamed resource:
  `"name":"postgresql_system_info"`

Closes #139

## Summary by CodeRabbit

* **New Features**
  * Built-in tools, resources, and prompts can now be enabled or disabled via environment variables, providing more flexible configuration options alongside the configuration file.

* **Documentation**
  * Updated configuration guides with comprehensive environment variable mappings for all built-in features.
  * Standardized naming conventions for built-in resources to ensure consistency.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -875,10 +875,26 @@ func setStringFromEnvWithFallback(dest *string, keys ...string) {
 
 // setBoolFromEnv sets a boolean config value from an environment variable if it exists
 // Accepts "true", "1", or "yes" (case-insensitive) as true values
+// parseBoolEnv interprets an environment-variable value as a boolean.
+// "true", "1", and "yes" are true; "false", "0", and "no" are false (all
+// case-insensitive). Any other value is treated as false, but a warning is
+// logged so operators get a diagnostic instead of a silent disable.
+func parseBoolEnv(key, val string) bool {
+	switch strings.ToLower(val) {
+	case "true", "1", "yes":
+		return true
+	case "false", "0", "no":
+		return false
+	default:
+		log.Printf("WARNING: unrecognised value %q for %s; expected "+
+			"true/false/1/0/yes/no — treating as false", val, key)
+		return false
+	}
+}
+
 func setBoolFromEnv(dest *bool, key string) {
 	if val := os.Getenv(key); val != "" {
-		lower := strings.ToLower(val)
-		*dest = lower == "true" || lower == "1" || lower == "yes"
+		*dest = parseBoolEnv(key, val)
 	}
 }
 
@@ -891,8 +907,7 @@ func setBoolPtrFromEnv(dest **bool, key string) {
 	if val == "" {
 		return
 	}
-	lower := strings.ToLower(val)
-	b := lower == "true" || lower == "1" || lower == "yes"
+	b := parseBoolEnv(key, val)
 	*dest = &b
 }
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -882,6 +882,20 @@ func setBoolFromEnv(dest *bool, key string) {
 	}
 }
 
+// setBoolPtrFromEnv sets a *bool config value from an environment variable if it exists.
+// Pointer-valued booleans distinguish "explicitly set" from "use default", so we only
+// allocate a new bool when the env var is present.
+// Accepts "true", "1", or "yes" (case-insensitive) as true values; everything else is false.
+func setBoolPtrFromEnv(dest **bool, key string) {
+	val := os.Getenv(key)
+	if val == "" {
+		return
+	}
+	lower := strings.ToLower(val)
+	b := lower == "true" || lower == "1" || lower == "yes"
+	*dest = &b
+}
+
 // setIntFromEnv sets an integer config value from an environment variable if it exists
 func setIntFromEnv(dest *int, key string) {
 	if val := os.Getenv(key); val != "" {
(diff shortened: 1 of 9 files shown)


internal/config/config.go
```diff
@@ -882,6 +882,20 @@ func setBoolFromEnv(dest *bool, key string) {
...
+// setBoolPtrFromEnv sets a *bool config value from an environment variable if it exists.
+// Pointer-valued booleans distinguish "explicitly set" from "use default", so we only
+// allocate a new bool when the env var is present.
+// Accepts "true", "1", or "yes" (case-insensitive) as true values; everything else is false.
+func setBoolPtrFromEnv(dest **bool, key string) {
+	val := os.Getenv(key)
+	if val == "" {
+		return
+	}
+	lower := strings.ToLower(val)
```
tsivaprasad: Emit a warning when an env var is set to an unrecognised value (e.g. `enabled`) so operators get a diagnostic instead of a silent disable.

Example:
```
	recognized := lower == "true" || lower == "1" || lower == "yes" ||
		lower == "false" || lower == "0" || lower == "no"
	if !recognized {
		log.Printf("WARNING: unrecognised value %q for %s; expected true/false/1/0/yes/no — treating as false", val, key)
	}
```

internal/config/config_test.go
```diff
@@ -813,6 +813,117 @@ func TestSetBoolFromEnv(t *testing.T) {
...
+			setBoolPtrFromEnv(&dest, key)
+			if dest == nil {
+				t.Fatalf("expected non-nil pointer for value %q", v)
+			}
+			if *dest {
+				t.Errorf("expected false for value %q, got true", v)
+			}
+		})
+	}
+
```
tsivaprasad, line 859: We can add one test here 
```
t.Run("unrecognised value is treated as false", func(t *testing.T) {
		os.Setenv(key, "enabled")
		defer os.Unsetenv(key)
		var dest *bool
		setBoolPtrFromEnv(&dest, key)
		if dest == nil {
			t.Fatal("expected non-nil pointer for unrecognised value")
		}
		if *dest {
			t.Errorf("expected false for unrecognised value, got true")
		}
	})
```

tsivaprasad (approved):
Looks good. I've requested one improvement.

internal/config/config.go
dpage: Done in 96ce00f. I pulled the shared bool parsing out of both `setBoolFromEnv` and `setBoolPtrFromEnv` into a `parseBoolEnv` helper that logs the warning, so the diagnostic covers both code paths rather than just one. Unrecognised values are still treated as false, but now emit: `WARNING: unrecognised value "enabled" for <KEY>; expected true/false/1/0/yes/no — treating as false`.

internal/config/config_test.go
dpage: Added in 96ce00f, as a subtest on `TestSetBoolPtrFromEnv` covering the unrecognised-value-treated-as-false contract. The full config suite passes.

PR #172, opened 2026-05-27
feat: replace hand-rolled LLM clients with pgedge-go-llm-lib (PR 1 of 4)

dpage:
## Summary

- Replaces hand-rolled Anthropic, OpenAI, and Ollama chat clients in `internal/chat/llm.go` with thin wrappers around [pgedge-go-llm-lib](https://github.com/pgEdge/pgedge-go-llm-lib).
- `chat.LLMClient` interface preserved; call sites in `internal/chat/client.go` and `internal/llmproxy/proxy.go` are essentially unchanged (only `NewOllamaClient` gained an `error` return).
- ~1500 lines of provider-specific HTTP code removed.

This is **PR 1 of 4** in the LLM library migration. Spec lives in `.claude/specs/2026-05-27-llm-library-migration-design.md` (gitignored / local-only).

Follow-up PRs:
- PR 2: Replace embedding providers (`internal/embedding/`).
- PR 3: Adopt library's HTTP proxy package; update web wire format.
- PR 4: Add streaming chat, add Gemini provider, delete shim layers.

## Known issues (filed upstream)

- Anthropic system prompt is no longer cached (pgedge-go-llm-lib#13).
- OpenAI Responses-API-only models not supported (pgedge-go-llm-lib#12).
- Pinned to a commit SHA pending v0.1.0 release (pgedge-go-llm-lib#14).

## Test plan

- [x] `make test` passes locally with no skips.
- [x] Manual live smoke test against real Anthropic API (`claude-sonnet-4-20250514`) returns expected response with correct token usage.
- [x] Manual live smoke test against real OpenAI API (`gpt-4o-mini`) returns expected response.
- [ ] Manual CLI smoke test against local Ollama — not run (no Ollama instance available in dev environment).
- [ ] Web client end-to-end smoke — deferred to PR 3 where the proxy wire format changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Upgraded the Go toolchain to 1.26 across CI, Docker builds, and the module directive.
  * Expanded CI and documentation build triggers to include `feat/llm-lib-*` branches.
* **Refactor**
  * Unified Anthropic, OpenAI, and Ollama LLM provider calls via a shared library.
  * Added OpenAI Responses API routing for `gpt-5-*`, `o1-*`, and `o3-*` models.
* **Bug Fixes**
  * Improved error handling when LLM client initialization fails (notably for Ollama).
* **Documentation**
  * Updated changelog entries for shared LLM client usage, broader Anthropic prompt caching, and OpenAI model support.
* **Tests**
  * Expanded LLM client tests, including tracing and message translation coverage.


dpage:
@coderabbitai review

dpage:
@coderabbitai review

dpage:
@coderabbitai review

internal/chat/llm_translate.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+			blocks = append(blocks, llm.ContentBlock{
+				Type:      llm.BlockToolResult,
+				ToolUseID: r.ToolUseID,
+				Text:      text,
+				IsError:   r.IsError,
+			})
+		}
+		return blocks, true, nil
+
```
tsivaprasad, line 90: contentToBlocks missing `[]interface{}` case — breaks all multi-tool conversations

internal/chat/llm_test.go
```diff
@@ -13,1012 +13,360 @@ package chat
...
+		t.Fatalf("Chat: %v", err)
+	}
+	if len(resp.Content) != 1 {
+		t.Fatalf("expected 1 content item, got %d", len(resp.Content))
+	}
+	tc, ok := resp.Content[0].(TextContent)
+	if !ok || tc.Text != "hi from ollama" {
+		t.Errorf("content[0] = %+v", resp.Content[0])
 	}
 }
```
tsivaprasad, line 477: `TestLibClient_Chat_MultiTurnToolLoop` is missing. 
Every existing test is single-turn — one Chat call, check the response. None of them exercise the agentic loop pattern where the first response is stored as history and passed back into the next call.

tsivaprasad (changes requested):

internal/chat/llm_translate.go
dpage: Fixed in b77e76c. I've added the missing `[]interface{}` case to `contentToBlocks`, dispatching each element through a new `elementToBlock` helper (covering `string`, `TextContent`, `ToolUse`, and `ToolResult`) and promoting the enclosing message to `RoleTool` when a tool-result element is present.

You were right about the impact: `[]interface{}` is exactly the shape an assistant turn takes once it has been stored back into the conversation history, because `LLMResponse.Content` is `[]interface{}` (see `fromLibContent`) and the agentic loop in `processQuery` appends it verbatim. Every multi-turn or multi-tool conversation therefore replayed that stored turn through `contentToBlocks`, hit the `default` branch, and failed with `unsupported content type []interface {}`.

The fix is merged forward through pr2/pr3/pr4; pr5 and pr6 delete this translation layer entirely (the wrapper is gone and the client uses `llmlib` types directly), so it does not apply there.

internal/chat/llm_test.go
dpage: Added in b77e76c. `TestLibClient_Chat_MultiTurnToolLoop` drives the full agentic loop against a fake provider: turn one returns a `tool_use` response, that response is stored as assistant history, a tool result is appended, and both are replayed into a second `Chat` call. It then asserts the replayed request carries the prior `tool_use` and `tool_result`, so it genuinely exercises the history round-trip rather than just a single call.

I've also added focused unit tests for the `[]interface{}` translation path directly (`TestToLibMessages_InterfaceSliceContent`, the tool-result role-promotion case, and an unknown-element error case). All of them fail with the original `unsupported content type []interface {}` error when the fix is reverted, which confirms they actually catch the regression.

PR #182, opened 2026-06-24
fix(installer): resolve latest stable v* release, not GitHub "latest"

AntTheLimey:
## Problem

The `examples/claude-try-it` install scripts trusted GitHub's `/releases/latest` and used its `tag_name` verbatim to build the binary download URL.

This repo also publishes **knowledge-base releases** (`kb-YYYY-MM-DD`) via `docker-publish.yml` that ship only a `kb.db` asset — no platform binaries. They are **not** flagged as prereleases, so `/releases/latest` returns them and the binary download 404s.

**Impact:** install/update has been broken for every user since **2026-04-22**, when the first `kb-*` release became "latest" (`v1.0.0`, the last release with binaries, shipped 2026-03-27).

```
ℹ  Downloading pgEdge MCP Server kb-2026-04-27 (darwin/arm64)...
curl: (56) The requested URL returned error: 404
✗  Download failed. Check your internet connection.
```

## Fix (client side)

- Fetch the releases **list** instead of `/releases/latest`.
- Keep stable `v*` releases only (drop `kb-*` and prerelease `-` tags).
- Order by **semantic version** (not GitHub's date order) so a hotfix on an older line can't be picked ahead of a newer release.
- Install the highest version that **actually ships a binary** for the current platform; reject a candidate only on a definitive **404** so a transient error / HEAD-blocking proxy can't strand users.
- Download failures now report the **HTTP status**: 403/429 explain the GitHub rate limit (and suggest pinning), 404 names the missing asset.
- New **version pin** escape hatch: `--version=` flag (`-Version` on Windows), precedence over the existing `PGEDGE_MCP_VERSION` env var. Documented in the example README.

Mirrored in `install.sh` and `install.ps1`.

## Testing

- `bash -n` clean; functional tests resolve `v1.0.0` against the live API; `kb-2026-04-27` correctly rejected (404).
- Semver sort verified (`v1.10.0` > `v1.9.0` regardless of list order).
- `--version=` flag + env-var precedence verified end-to-end.
- 404 download path verified to emit the new message and exit non-zero.
- PowerShell logic mirrors bash but was **not** run locally (no `pwsh` available) — worth a quick check on Windows before merge.

## Follow-ups (not in this PR)

- **Layer 2 (defense in depth):** mark `kb-*` releases as `prerelease: true` / `make_latest: false` in `docker-publish.yml` so `/releases/latest` is correct too.
- Cosmetic: `ask()` errors `"/dev/tty: Device not configured"` in non-TTY/CI contexts; make the TTY probe check readability.

## Summary by CodeRabbit

* **New Features**
  * Added `--version` (macOS/Linux) and `-Version` (Windows) options to pin the pgEdge MCP Server release during installation.
  * Version pinning now takes precedence over `PGEDGE_MCP_VERSION` and avoids “latest” discovery.
* **Bug Fixes**
  * Improved download/version resolution with better validation of platform-specific binaries.
  * Enhanced error handling and cleanup, including clearer messages for rate limits and missing binaries.
* **Documentation**
  * Added a “Pinning a Version” section with bash and PowerShell examples.

### DIFF
--- a/examples/claude-try-it/install.ps1
+++ b/examples/claude-try-it/install.ps1
@@ -102,16 +102,19 @@ function Get-Platform {
 # --- Get latest release version -------------------------------------------
 
 # Does the server binary asset exist for a given tag on this platform?
-# Rejects only on a definitive 404 — any other failure (transient network, a
-# proxy that blocks HEAD) is treated as "present" so we don't wrongly skip a
-# good release and strand every Windows user.
+# Uses a 0-byte ranged GET (Range: bytes=0-0), matching install.sh: HEAD is
+# unreliable against GitHub's asset redirect to S3, which can reject it. Rejects
+# only on a definitive 404 — any other failure (transient network, a proxy
+# quirk) is treated as "present" so we don't wrongly skip a good release and
+# strand every Windows user.
 function Test-AssetExists {
     param([string]$Tag)
     $num = $Tag.TrimStart('v')
     $asset = "pgedge-postgres-mcp-server_${num}_$($script:OS)_$($script:Arch).$($script:Ext)"
     $url = "https://github.com/$Repo/releases/download/$Tag/$asset"
     try {
-        Invoke-WebRequest -Uri $url -Method Head -UseBasicParsing -ErrorAction Stop | Out-Null
+        Invoke-WebRequest -Uri $url -Headers @{ Range = "bytes=0-0" } `
+            -UseBasicParsing -ErrorAction Stop | Out-Null
         return $true
     } catch {
         if ($_.Exception.Response -and [int]$_.Exception.Response.StatusCode -eq 404) {
@@ -151,8 +154,14 @@ function Get-LatestVersion {
     # Stable v* releases only, ordered by semantic version (highest first).
     # GitHub lists releases by date, not version, so a hotfix cut on an older
     # line after a newer release would otherwise be picked ahead of it.
+    #
+    # Filter structurally on the tag, matching install.sh: reject any tag with a
+    # "-" (v1.0.0-beta3, -rc1…) rather than trusting GitHub's prerelease flag.
+    # This repo does not set that flag reliably — its beta/alpha tags are all
+    # published with prerelease=false — so the flag alone would let an RC slip
+    # through whenever a release line has no stable counterpart yet.
     $candidates = @($releases |
-        Where-Object { -not $_.draft -and -not $_.prerelease -and $_.tag_name -match '^v\d' } |
+        Where-Object { -not $_.draft -and -not $_.prerelease -and $_.tag_name -match '^v\d' -and $_.tag_name -notmatch '-' } |
         ForEach-Object { $_.tag_name } |
         Sort-Object { try { [version]($_.TrimStart('v')) } catch { [version]'0.0' } } -Descending)
 
--- a/examples/claude-try-it/install.ps1
+++ b/examples/claude-try-it/install.ps1
@@ -12,6 +12,10 @@
 #   .\install.ps1 -Detect -DbPort 5433
 #   .\install.ps1 -OwnDb -DbHost localhost -DbPort 5432 -DbName mydb -DbUser me -DbPass secret
 #
+# Pin a specific version (escape hatch — skips the latest-release lookup):
+#   .\install.ps1 -Version v1.0.0
+#   $env:PGEDGE_MCP_VERSION = "v1.0.0"; irm .../install.ps1 | iex   # env var equivalent
+#
 # What it does:
 #   1. Downloads the pgEdge MCP Server binary for Windows (x86_64)
 #   2. Helps you connect to a database (your own or a demo with sample data)
@@ -26,7 +30,8 @@ param(
     [string]$DbPort,
     [string]$DbName,
     [string]$DbUser,
-    [string]$DbPass
+    [string]$DbPass,
+    [string]$Version
 )
 
 $ErrorActionPreference = 'Stop'
@@ -44,6 +49,8 @@ $DemoPort   = 5432
 
 $script:Version      = ""
 $script:VersionNum   = ""
+# Pinned version: -Version flag wins, else PGEDGE_MCP_VERSION env var, else empty.
+$script:PinVersion   = if ($Version) { $Version } elseif ($env:PGEDGE_MCP_VERSION) { $env:PGEDGE_MCP_VERSION } else { "" }
 $script:DbHost       = $DbHost
 $script:DbPort       = $DbPort
 $script:DbName       = $DbName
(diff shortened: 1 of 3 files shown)


examples/claude-try-it/install.sh
```diff
@@ -99,13 +106,78 @@ detect_platform() {
...
+# Instead, walk the releases list newest-first and pick the first STABLE
+# software release (tag matches v<digit>… with no prerelease "-" suffix) that
+# actually has a server binary for this platform. A pinned version (--version=
+# flag or PGEDGE_MCP_VERSION env var) overrides everything as an escape hatch
+# (e.g. --version=v1.0.0).
 get_latest_version() {
+  if [ -n "$PIN_VERSION" ]; then
+    VERSION="$PIN_VERSION"
+    VERSION_NUM="${VERSION#v}"
+    info "Using pinned version $VERSION"
```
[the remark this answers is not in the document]
AntTheLimey: Fixed in 211373c. The pin now normalizes to a `v`-prefixed tag — `VERSION_NUM="${PIN_VERSION#[vV]}"; VERSION="v${VERSION_NUM}"` — so `--version=1.0.0`, `v1.0.0`, and `V1.0.0` all resolve to the `v1.0.0` tag, matching the README. Verified end-to-end: `--version=1.0.0` → `Using pinned version v1.0.0`.

examples/claude-try-it/install.ps1
```diff
@@ -94,17 +101,72 @@ function Get-Platform {
...
+#
+# Instead, walk the releases list newest-first and pick the first STABLE
+# software release (v<digit>…, not draft, not prerelease) that actually has a
+# server binary for this platform. A pinned version (-Version flag or
+# PGEDGE_MCP_VERSION env var) overrides everything as an escape hatch.
 function Get-LatestVersion {
+    if ($script:PinVersion) {
+        $script:Version = $script:PinVersion
+        $script:VersionNum = $script:Version.TrimStart('v')
+        Write-Info "Using pinned version $($script:Version)"
```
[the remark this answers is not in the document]
AntTheLimey: Fixed in 211373c, same normalization on the PowerShell side: `$script:VersionNum = $script:PinVersion.Trim().TrimStart('v','V'); $script:Version = "v$($script:VersionNum)"`. `-Version 1.0.0` now resolves to `v1.0.0`.

examples/claude-try-it/install.ps1
```diff
@@ -120,7 +182,15 @@ function Install-Binary {
...
-        Write-Fail "Download failed. Check your internet connection."
+        $code = if ($_.Exception.Response) { [int]$_.Exception.Response.StatusCode } else { 0 }
+        if ($code -eq 403 -or $code -eq 429) {
+            Write-Fail "Download failed (HTTP $code): GitHub rate limit reached. Wait a few minutes and retry, or pin a version with -Version $($script:Version)."
+        } elseif ($code -eq 404) {
+            Write-Fail "Download failed (HTTP 404): no $($script:OS)/$($script:Arch) binary for $($script:Version) at $url"
+        } else {
+            $suffix = if ($code) { " (HTTP $code)" } else { "" }
+            Write-Fail "Download failed$suffix. Check your internet connection."
+        }
```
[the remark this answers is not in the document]
AntTheLimey: Fixed in 211373c. Added `Remove-Item $tmpDir -Recurse -Force -ErrorAction SilentlyContinue` in the catch block before the `Write-Fail` branches, so the temp dir is cleaned up on all error paths (this matches the `rm -rf "$tmp_dir"` already done before each `fail` in install.sh).

examples/claude-try-it/install.ps1
```diff
@@ -94,17 +101,74 @@ function Get-Platform {
...
+        Write-Fail "Could not fetch releases from GitHub: $_"
     }
-    if (-not $script:Version) {
-        Write-Fail "Could not determine latest release version"
+
+    # Stable v* releases only, ordered by semantic version (highest first).
+    # GitHub lists releases by date, not version, so a hotfix cut on an older
+    # line after a newer release would otherwise be picked ahead of it.
+    $candidates = @($releases |
+        Where-Object { -not $_.draft -and -not $_.prerelease -and $_.tag_name -match '^v\d' } |
```
dpage: 🟡 **Should Fix — PowerShell detects pre-releases differently from `install.sh`, and the strategy is wrong for this repo's tagging convention.**

Here PowerShell trusts GitHub's `prerelease` boolean, whereas `install.sh:154` filters structurally on the `-` in the tag (`case "$tag" in *-*) continue`). The problem is that this repo does **not** set the `prerelease` flag reliably: the live release list shows `v1.0.0-beta3`, `-beta2`, `-beta1`, `-alpha6`, `-alpha5` all with `prerelease=false` (only `v1.0.0-test1` is flagged `true`). So this filter will accept a `v*-beta/-rc` tag as a stable candidate.

Today it survives by luck — `[version]('1.0.0-beta3')` throws, the `catch` maps it to `0.0`, and the genuine `v1.0.0` sorts above it. But the first time a release line has only a pre-release tag and no stable counterpart (e.g. early in a `v1.1.0` cycle, `v1.1.0-rc1` left at `prerelease=false`), bash correctly reports "no stable release" whilst PowerShell would install the RC the user never asked for. The whole point of this PR is to stop trusting GitHub's release metadata; the same scepticism should apply to the `prerelease` flag.

Suggest matching the bash behaviour:
```powershell
Where-Object { -not $_.draft -and -not $_.prerelease -and $_.tag_name -match '^v\d' -and $_.tag_name -notmatch '-' }
```

examples/claude-try-it/install.ps1
```diff
@@ -94,17 +101,74 @@ function Get-Platform {
...
+# Rejects only on a definitive 404 — any other failure (transient network, a
+# proxy that blocks HEAD) is treated as "present" so we don't wrongly skip a
+# good release and strand every Windows user.
+function Test-AssetExists {
+    param([string]$Tag)
+    $num = $Tag.TrimStart('v')
+    $asset = "pgedge-postgres-mcp-server_${num}_$($script:OS)_$($script:Arch).$($script:Ext)"
+    $url = "https://github.com/$Repo/releases/download/$Tag/$asset"
+    try {
+        Invoke-WebRequest -Uri $url -Method Head -UseBasicParsing -ErrorAction Stop | Out-Null
```
dpage: ℹ️ **Note — asset-existence probe uses a different HTTP method from the bash version.** `asset_exists_for_version` in `install.sh` deliberately uses a `0-0` ranged GET, with a comment explaining that HEAD is unreliable against GitHub's redirect to S3; here `Test-AssetExists` still uses `-Method Head`. If that HEAD is blocked it falls back to "present" (the safe default), so users aren't stranded — the only consequence is the Windows asset-skip safety net may silently not skip. Impact is low (every `v*` release ships the Windows binary, and the primary `kb-*` bug is cured by the `v*` filter regardless), but worth either a comment noting the divergence is intentional or aligning onto the same ranged-GET technique for parity.

examples/claude-try-it/install.sh
```diff
@@ -99,13 +106,80 @@ detect_platform() {
...
+    return
+  fi
+
   local response
-  response=$(curl -fsSL "https://api.github.com/repos/$REPO/releases/latest") \
-    || fail "Could not fetch latest release from GitHub (network error or rate limit)"
-  VERSION=$(echo "$response" | grep -o '"tag_name": *"[^"]*"' | head -1 | cut -d'"' -f4) || true
-  [ -z "$VERSION" ] && fail "Could not determine latest release version"
-  VERSION_NUM="${VERSION#v}"
+  response=$(curl -fsSL "https://api.github.com/repos/$REPO/releases?per_page=30") \
```
dpage, line 146: ℹ️ **Note — `per_page=30` is comfortable today but coupled to `kb-*` cadence.** If a release line ever has no stable `v*` tag within the 30 most recent releases (only plausible if `kb-*` releases became frequent), resolution fails with "No stable release found". Only two `kb-*` releases exist so far, so this is purely theoretical, and the documented Layer-2 follow-up (flagging `kb-*` as `prerelease`/`make_latest: false`) removes the risk entirely. No action needed now; just flagging the coupling. Same applies to `install.ps1:146`.

dpage (commented):
Reviewed the version-resolution rework. This is a well-reasoned, well-commented fix for a genuine and impactful bug, and the documentation is thorough. The three CodeRabbit findings are all addressed and resolved, so I won't repeat them. One design inconsistency is worth fixing before merge, plus a couple of lower-priority notes (left inline).

**Severity key:** 🟡 Should Fix · ℹ️ Note

examples/claude-try-it/install.ps1
AntTheLimey: Fixed in f48c55b. Confirmed the diagnosis against the live release list — `v1.0.0-beta3/-beta2/-beta1/-alpha6/-alpha5` are all published with `prerelease=false` (only `-test1` is flagged). Adopted your suggested filter so PowerShell now matches the bash structural check:

```powershell
Where-Object { -not $_.draft -and -not $_.prerelease -and $_.tag_name -match '^v\d' -and $_.tag_name -notmatch '-' }
```

So a `v*-rc/-beta` tag can no longer be picked as stable even on a line with no stable counterpart yet.

examples/claude-try-it/install.ps1
AntTheLimey: Aligned onto the ranged-GET technique in f48c55b rather than just documenting the divergence — `Test-AssetExists` now issues `Invoke-WebRequest -Headers @{ Range = "bytes=0-0" }` instead of `-Method Head`, matching `asset_exists_for_version` in install.sh. Same 404-only rejection semantics, so the Windows asset-skip safety net is now reliable rather than dependent on HEAD surviving the S3 redirect.

examples/claude-try-it/install.sh
AntTheLimey: Agreed — leaving as-is for now. With only two `kb-*` releases the 30-release window has comfortable headroom, and the Layer-2 follow-up (flagging `kb-*` as `prerelease`/`make_latest: false`) removes the coupling entirely. Noted for that follow-up.

dpage (approved):

PR #192, opened 2026-07-21
fix: detect secret rotation via atomically-swapped symlinks (#186)

### LINKED ISSUE #186: Token/config file watcher ignores atomic symlink replacement
The file watcher only reacts to `Write` and `Create` events on the exact watched path. 

When the file is delivered by something that swaps a symlink atomically (an orchestrator projecting a mounted secret, or any tool that renames into place), the events that fire are `Rename`/`Remove`/`Create` on the linked target, not `Write`/`Create`. 

The watcher misses those, so the file is never reloaded and updates only take effect on restart. 

Proposed fix: watch the parent directory and/or also handle `Rename` and `Chmod`, then re-resolve and re-read the target.

moizpgedge:
## Summary

- The token and user file watchers only reacted to `Write`/`Create` fsnotify events on the exact watched filename, so a secret delivered via an atomically-swapped symlink (a Kubernetes-projected Secret/ConfigMap volume, or any tool that renames a new version into place) was silently missed and only picked up on restart.
- Root cause: Kubernetes never touches the watched filename (`tokens.yaml`) directly — it's a symlink through a hidden `..data` symlink that gets atomically re-pointed at a new timestamped directory on rotation. The events that fire are `Create`/`Rename`/`Remove` on `..data` and a transient temp name, never on the watched path, so the old exact-name filter dropped every one of them regardless of event type.
- The fix: the watcher now reacts to *any* event in its watched directory, then re-resolves and hashes the watched path's own content (following symlinks transparently via `os.Open`) to decide whether a reload is actually warranted. This catches changes delivered through any indirection while still correctly ignoring unrelated activity elsewhere in the directory — covered by the existing `TestWatcherIgnoresOtherFiles` test, which still passes unmodified.

## Test plan

- [x] New `TestWatcherDetectsKubernetesSecretRotation` reproduces the real Kubernetes double-symlink rotation mechanism (populate a new versioned directory, atomically rename a temp symlink onto `..data`). Verified it fails against the pre-fix code (`count == 0`) and passes against the fix (`count == 1`).
- [x] All 9 watcher tests, including the 8 pre-existing ones, pass individually and as part of the full `internal/auth` suite under `-race`, matching `ci-server.yml`'s exact invocation. Repeated 5x back-to-back to rule out debounce/timing flakiness — no failures.
- [x] `-race` caught a real data race introduced mid-development (two debounced checks racing on the stored content hash); fixed with a mutex around the full check-compare-reload sequence and re-verified clean.
- [x] `make test-server` — full exit code (not a piped subcommand's) confirmed 0, zero `FAIL` across `internal/mcp`, `internal/auth`, `internal/config`, `internal/crypto`, `internal/database`, `internal/openapi`, `internal/resources`, `internal/tools`, `internal/tracing`.
- [x] `make lint-server` — 0 issues.
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` — all clean.
- [x] `docs/changelog.md` updated under `[Unreleased] / Fixed`.

Closes #186

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication/token and user configuration reloads are now reliably triggered when files update via atomic symlink/swap mechanisms (including Kubernetes-style projected secrets/configs).
  * Reload checks are more dependable, avoiding missed updates during directory-level event bursts.
  * Stopping the watcher is safer by waiting for any in-flight reload to finish and supporting repeated Stop calls.
* **Documentation**
  * Added an **Unreleased → Fixed** changelog entry describing the improved projected-update reload behavior.
* **Tests**
  * Added regression tests for projected secret rotation, in-flight reload stopping, and idempotent/multiple Stop calls.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -149,6 +149,18 @@ and this project adheres to
 
 ### Fixed
 
+- The token and user file watchers now detect changes delivered through an
+  atomically-swapped symlink, such as a Kubernetes-projected Secret or
+  ConfigMap volume, or any tool that renames a new version into place.
+  Previously the watcher matched events by exact filename and only handled
+  `Write`/`Create`, so a symlink swap on a different directory entry (for
+  example Kubernetes' own `..data` symlink) never triggered a reload and
+  updates only took effect on restart. The watcher now reacts to any event
+  in the watched directory and re-resolves and hashes the watched path's
+  content to decide whether a reload is warranted, catching changes that
+  never touch the watched filename directly while still ignoring
+  unrelated activity elsewhere in the directory. (#186)
+
 - Metadata loader now tolerates tables with zero columns
   (e.g. `CREATE TABLE foo()`). The query LEFT JOINs against the
   per-column catalog, so a zero-column table produced a row whose
--- a/internal/auth/watcher.go
+++ b/internal/auth/watcher.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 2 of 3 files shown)


dpage (approved):
Approving after review of the atomic-symlink watcher fix. Root cause and approach are sound, shutdown handling is correct, tests reproduce the real Kubernetes double-symlink rotation and pass under -race. Also pushed a follow-up so a failed reload leaves the content hash unchanged and is retried on the next event rather than stranding stale data. All CI checks green.

PR #193, opened 2026-07-21
fix: show configured display name instead of raw connection details (…

### LINKED ISSUE #187: Tool responses leak the raw connection DSN instead of a display name
Several tool responses (system-info, and some schema-info and query paths) return the full connection string including the database host instead of the configured display name. 

When the host is an internal-only address the caller can't reach, this exposes infrastructure detail and confuses clients. 

The display-name substitution exists but isn't applied everywhere.

Proposed fix: use the configured display name in every user-facing string, or add a `display_host`/`display_url` field that always renders in place of the DSN.

moizpgedge:
## Summary

- Tool and resource responses only masked the password in the connection string they showed the caller, leaving the real host, port, and database name fully visible. `pg://system_info` was worse: it queried `inet_server_addr()`/`inet_server_port()` live, reporting the actual server-side network address of the connection rather than anything the operator configured.
- Confirmed live against a real Postgres container: connecting via its Docker DNS name returned the container's internal bridge IP (`172.19.0.2`, unreachable from outside that network) as `"host"` in the resource, and the same raw address leaked through `query_database`/`get_schema_info`/`execute_explain`/`count_rows`/`similarity_search` — with the operator's configured display name appearing nowhere.
- `Client.DisplayName()` now backs every one of these responses with the connection's configured `name`, falling back to a password-masked connection string when none is configured. Two new getters, `ConfiguredHost()`/`ConfiguredPort()`, give `pg_system_info` the operator's configured address (matching what `list_database_connections`/`GET /api/databases` already expose) instead of a live-resolved one; the resource also gains a `connection_name` field carrying the display name directly.
- Deliberately left alone: the ad-hoc `postgres://...` connection strings a caller can type inline in `query_database`'s mini-DSL. Echoing back what the caller themselves supplied isn't a leak, so those paths still show a password-masked version of their own literal input rather than a display name.

## Test plan

- [x] New unit tests for `DisplayName`/`ConfiguredHost`/`ConfiguredPort` cover configured name, no config, empty name, multi-host failover configs, and port defaulting — pure logic, no DB required.
- [x] New `TestGetSchemaInfoTool_UsesDisplayName` and `TestPGSystemInfoResource_UsesConfiguredDisplayValues` reproduce the bug precisely (a distinctive internal-looking host that must never appear in the response). Confirmed non-tautological: reverted against the pre-fix code, the `get_schema_info` test fails exactly as expected, and the `system_info` test fails to even compile since `connection_name` doesn't exist without the fix.
- [x] The `system_info` test was run for real against a live local PostgreSQL instance, proving the fix end-to-end through actual SQL execution, not mocked logic.
- [x] Rebuilt the actual server binary and drove it through the real MCP JSON-RPC protocol (`initialize`, `resources/read`, `tools/call`) against a live Postgres container, confirming `query_database`, `get_schema_info`, `execute_explain`, `count_rows`, and `pg://system_info` all now show the configured display name end-to-end.
- [x] Full `internal/database`, `internal/resources`, `internal/tools` suites pass under `-race`, both with and without a live `TEST_PGEDGE_POSTGRES_CONNECTION_STRING`. One unrelated pre-existing failure (`TestClientManager_Concurrency`) was confirmed to fail identically on unmodified `main` with the same environment — not a regression from this change.
- [x] `make lint-server`: 0 issues. `make test-server`: exit code 0 (verified directly, not through a piped `tail`), zero failures across every server package.

Closes #187

## Summary by CodeRabbit

* **Bug Fixes**
  * Database tools now show the operator-configured database display name instead of raw connection details (host/port), with passwords still masked when no name is available.
  * `pg://system_info` now reports configured host/port values and includes an optional `connection_name`.
* **Documentation**
  * Updated the changelog and system information reference to document the revised connection details and the new `connection_name` field.
* **Tests**
  * Added regression tests covering display name, configured host/port selection, and `pg://system_info` output.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -149,6 +149,25 @@ and this project adheres to
 
 ### Fixed
 
+- Tool and resource responses now show the operator-configured database
+  display name instead of the raw connection details. Previously,
+  `query_database`, `get_schema_info`, `execute_explain`, `count_rows`,
+  and `similarity_search` only masked the password in the connection
+  string they showed the caller, leaving the real host, port, and
+  database name visible; `pg://system_info` was worse, reporting the
+  live-resolved server address from `inet_server_addr()`, which can be
+  an internal-only address (a container or pod IP) that differs from,
+  and may be unreachable via, the address the operator actually
+  configured. A new `Client.DisplayName()` now backs every one of
+  these responses with the connection's configured `name` (falling
+  back to a password-masked connection string when none is
+  configured); `pg://system_info` gains a `connection_name` field and
+  its `host`/`port` fields now reflect the configured values rather
+  than a live-resolved one. Ad-hoc connection strings a caller types
+  inline (the `postgres://...` mini-DSL supported by `query_database`)
+  are intentionally left as-is, since echoing back what the caller
+  themselves supplied is not a leak. (#187)
+
 - Metadata loader now tolerates tables with zero columns
   (e.g. `CREATE TABLE foo()`). The query LEFT JOINs against the
   per-column catalog, so a zero-column table produced a row whose
--- a/docs/reference/resources.md
+++ b/docs/reference/resources.md
@@ -90,7 +90,7 @@ When you read the resource to view PostgreSQL system information, the result is 
 | `user` | Current database user. |
 | `host` | The operator-configured connection host (or `unix socket`), never a live-resolved server address. |
 | `port` | The operator-configured connection port. |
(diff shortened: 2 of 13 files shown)


dpage (approved):
Reviewed in depth: clean, well-scoped fix for #187. Verified the display-name substitution is applied on every user-facing path (no raw configured host leaks remain), the inline-vs-configured branch in query_database is correct, and pg://system_info now reports operator-configured host/port/name rather than the live-resolved server address. Tests are thorough and non-tautological; full suite and CI are green. All three CodeRabbit findings addressed. LGTM.

PR #196, opened 2026-07-28
Harden the read-only guardrails against transaction-mode bypass

dpage:
## Background

A security report showed that read-only mode could be escaped using ordinary SQL transaction forms, verified independently of the MCP server by creating, modifying and dropping real tables. Tracing it found two contributing causes.

The textual guard matched only the literal strings `transaction_read_only` and `default_transaction_read_only`, and never `READ WRITE` in any form. Separately, every statement that did not begin with `SELECT`, `WITH`, `TABLE` or `VALUES` was executed through pgx's `Exec`, which falls back to the PostgreSQL simple query protocol whenever no bind parameters are supplied (`conn.go:500`, "Always use simple protocol when there are no arguments", and the override sits after the option loop, so a mode argument cannot opt out of it). That protocol accepts any number of semicolon-separated statements per message, so a caller could append their own statements, request a read-write access mode or commit the server's own transaction, and then write.

Worth emphasising: a leading comment alone was enough to reach that path, so the bypass never needed a keyword the guard was looking for. Filtering harder would not have fixed this.

## What this changes

The structural fix is to stop offering the channel. On a read-only connection every statement now runs through `tx.Query`, and therefore the extended query protocol, which carries exactly one statement per message and has PostgreSQL reject anything else. Connections configured with `allow_writes: true` keep the previous behaviour, since there is no boundary left to protect there and multi-statement scripts remain useful.

Alongside that:

- All four tools request read-only as part of `BEGIN` via `BeginTx`, instead of issuing `SET TRANSACTION READ ONLY` afterwards, so the transaction is never briefly writable and the mode cannot fail to apply separately.
- The guard now covers the read-write access modes, `SET SESSION CHARACTERISTICS`, `RESET ALL`, `DISCARD`, transaction control, `SET ROLE`, `SET SESSION AUTHORIZATION`, `ALTER ROLE`/`USER`/`DATABASE`, and operations acting outside the transaction that read-only mode never restrained at all: `DO` blocks, `COPY ... TO PROGRAM`, the server-side file functions, and `dblink`. Matching runs after comments and literals are stripped, so a comment cannot split a keyword, and a query that merely mentions a setting name is no longer refused. It now also screens the `count_rows` `where` argument and the `execute_explain` query.
- The session-level default is re-applied when a pooled connection is released, and the connection is discarded if that cannot be confirmed. Previously a `RESET ALL` or `DISCARD ALL` persisted on that pooled connection for every later caller of it. **Corrected since first posting:** this was not a route through `query_database`, which set the access mode on each of its own transactions and so stayed protected; it mattered for paths that did not, such as the `pl-func` custom tool executor, and it disabled the session-level layer for the rest of that connection's life. Testing against unmodified `main` is what established the narrower scope.
- Custom `pl-do` tools no longer interpolate arguments between fixed dollar-quote delimiters. JSON encoding does not escape a dollar sign, so an argument containing `$mcp_custom_tool$` or `$mcp_args$` closed the quoting and had the remainder of its value executed; combined with the simple query protocol used for those statements, that was a second full bypass that the guard never saw. Delimiters are now generated per invocation, for `pl-func` as well.
- `pl-func` tools are refused up front on a read-only connection instead of failing partway through with an opaque permissions error, which invited disabling read-only mode as the remedy.
- Rejected statements are logged in full, since a rejection is evidence of an attempted escape and nothing previously recorded what was tried. Note that this means statement text, and so possibly data values, reaches the log.

## What this does not change

None of these layers is a genuine constraint: each is a setting the connected role is entitled to change, and the guard can only ever reject what it recognises. The documentation now says so plainly, and points at the one control a client cannot undo, which is a role with writes revoked at the privilege level. `docs/guide/security.md` gains a section on the limits, and `docs/guide/security_mgmt.md` notes the primitives that privileges alone do not restrain either.

## Testing

**Update: the bypass is now reproduced and the fix verified against a live PostgreSQL 18.4.** My first posting said this was unverified for want of a database; I since built a throwaway cluster and ran `TestReadOnlyStatementSmuggling` on both sides.

Against unmodified `main`, two payloads create real tables:

```
--- FAIL: TestReadOnlyStatementSmuggling
    SET_TRANSACTION_READ_WRITE_then_write: read-only bypassed: smuggled CREATE TABLE succeeded
    COMMIT_then_BEGIN_READ_WRITE_then_write: read-only bypassed: smuggled CREATE TABLE succeeded
```

Against this branch, all five payloads fail to write and legitimate queries still work:

```
--- PASS: TestReadOnlyStatementSmuggling (0.52s)
ok      pgedge-postgres-mcp/test
```

So the reported bypass is real, and this branch closes it. Two details worth recording from that run. The leading-comment payload reaches the vulnerable `Exec` path but does not by itself complete a write, since the per-transaction `SET TRANSACTION READ ONLY` still applied: it is the route in, not the whole exploit. And the `RESET ALL` and `SET SESSION CHARACTERISTICS` payloads did not bypass `main` either, for the reason given in the corrected bullet above, which is why the pool-poisoning claim has been narrowed.

### Suite status

`make test` passes with no failures, `make lint` reports 0 issues, and `gofmt` and `go vet` are clean. The guard has 46 table-driven cases, plus unit tests for the literal and comment scanner and a regression test for the `pl-do` delimiter escape.

**The live exploit has not been re-run.** No PostgreSQL credentials were available in the environment where this was written, so the original bypass was neither reproduced nor observed failing. That proof is written as `TestReadOnlyStatementSmuggling` in `test/integration_test.go`, gated on `TEST_PGEDGE_POSTGRES_CONNECTION_STRING`: it fires five payloads, including the two from the report and the leading-comment variant, then checks out-of-band that no table was created, and skips itself if the configured role could not create one anyway so that a pass cannot be vacuous. Please run it against a real database before treating the bypass as closed.

Two further notes for review. The guard will occasionally over-block, for instance a query using `discard`, or both `copy` and `program`, as unquoted identifiers; that seemed the right direction to fail for a security control. And statements that previously took the simple protocol are now prepared, which matters only for PgBouncer in transaction pooling mode, and is not a change in kind because every `SELECT` already used the extended protocol.

## Summary by CodeRabbit

* **Security**
  * Strengthened read-only enforcement by starting transactions in read-only access mode at `BEGIN` and revalidating pooled connections to prevent writable session state from leaking.
  * Expanded read-only guardrails to block write/escape vectors, statement smuggling, and disallowed session/transaction changes; improved behavior so `transaction_read_only` mentions inside literals/comments don’t trigger false rejections.
  * Hardened custom tools: `pl-do`/`pl-func` wrapping now resists delimiter-based bypasses, and `pl-func` execution is refused when writes aren’t permitted.
* **Documentation**
  * Reworked read-only “defense-in-depth” guidance and deployment recommendations.
* **Tests**
  * Added end-to-end read-only enforcement coverage and expanded read-only validation/unit/regression tests.


dpage:
Verified against a live database since posting, using a throwaway PostgreSQL 18.4 cluster.

**The bypass reproduces on `main`.** Two of the five payloads create real tables there:

- `SET TRANSACTION READ WRITE; CREATE TABLE ...`
- `COMMIT; BEGIN READ WRITE; CREATE TABLE ...; COMMIT`

**All five fail on this branch**, and ordinary queries continue to work. So the report was accurate and this change closes it.

Two corrections came out of that run, both now reflected in the description and the changelog:

1. The pool-poisoning claim was too strong. `RESET ALL` and `SET SESSION CHARACTERISTICS AS TRANSACTION READ WRITE` did **not** bypass `main`, because `query_database` set the access mode on every one of its own transactions and so stayed protected after the session default was cleared. Poisoning is real but narrower: it disables the session-level layer for the rest of that connection's life and matters for paths that do not set the mode themselves, notably the `pl-func` custom tool executor. The re-assertion on release remains worth having as defence in depth, which is now what it claims to be.

2. The leading-comment payload reaches the vulnerable `Exec` path but does not on its own complete a write, since the per-transaction mode still applied. It is the route in rather than the whole exploit, which is how the description already framed it, but worth stating explicitly now there is evidence either way.

dpage:
Added end-to-end coverage for the custom tools framework, prompted by a report that the `pl-func` tool type writes to the schema even when `allow_writes: false`.

**That claim does not hold.** Run against unmodified `main` with a live PostgreSQL 18.4, the tool answers:

```
failed to create function: ERROR: cannot execute CREATE FUNCTION in a read-only transaction (SQLSTATE 25006)
```

and `pg_proc` shows nothing created. The report correctly spotted that `executePLFuncInTransaction` never issues `SET TRANSACTION READ ONLY`, but the session-level `default_transaction_read_only` applied in `AfterConnect` already made the transaction read-only, so the write cannot happen. The `pl-func` type was **broken** under `allow_writes: false`, not a bypass.

What was genuinely wrong is narrower and is what this branch fixes: that path depended on a single guardrail, and it failed opaquely partway through in a way that invites an operator to reach for `allow_writes` as the remedy. It now refuses up front and names the setting.

The report's more valuable observation was that the framework had **no end-to-end coverage at all**, because it is invisible until an operator sets `custom_definitions_path` — which is exactly how a tool type came to rest on one guardrail unnoticed. `TestCustomToolsRespectReadOnly` now enables it the way an operator would and covers all three types against a read-only connection:

| Case | Result |
|---|---|
| `sql` tool, read | succeeds |
| `sql` tool, write | refused, `cannot execute INSERT in a read-only transaction` |
| `pl-do` tool, read | succeeds |
| `pl-do` tool, write | refused, same error, so procedural code gets no exemption |
| `pl-func` tool | refused up front, naming `allow_writes` |

Each case also asserts out of band that the fixture table is unchanged and that no `_mcp_custom_tool_%` function survives in `pg_proc`. Against `main` the only failing assertion is the one about the message naming the setting, which is precisely the improvement here.

moizpgedge (approved):

PR #208, opened 2026-07-30
fix(security): refresh postgres-mcp Trivy ignore list

moizpgedge:
## What this fixes

A security scan of the published server image flagged 2 Critical and
46 High-severity warnings. Almost all of them came down to one thing:
the published image was old. It was built before several software
updates that had already happened. Rebuilding it from the current
version, with no other changes, fixed nearly everything on the list.

## What was left

A handful of warnings remained in a small helper library used only to
download one file during setup (no fix is available for these yet).
I checked what each one actually requires to be a problem, and none of
those conditions exist here: this download happens once, with no login,
no special network setup, and nothing an attacker could realistically
reach. These are documented as "checked, doesn't apply" rather than
silently ignored.

There were also about 113 lower-severity warnings, all in software
included in the base image but never actually used by the server
itself. Same approach: wrote a specific, checkable reason for each one,
then double-checked the entire list against a fresh scan to make sure
nothing was missed or left over from before.

## Results

Scanning the newly built image against the updated list of "checked,
doesn't apply" items shows zero warnings, at every severity level.

## What this doesn't do

happens.
https://pgedge.atlassian.net/browse/PLAT-701

## Summary by CodeRabbit

* **Chores**
  * Updated vulnerability exceptions for the Postgres MCP server’s base-image packages.
  * Refined vulnerability assessments to reflect the server’s runtime behavior and non-exploitable code paths.
  * Added package identification and expiration metadata to improve tracking and lifecycle management.
  * Removed outdated exceptions and added newer relevant vulnerability entries.

### DIFF
--- a/.trivy/postgres-mcp.trivyignore.yaml
+++ b/.trivy/postgres-mcp.trivyignore.yaml
@@ -74,17 +74,62 @@ vulnerabilities:
       Integer overflow DoS in g_buffered_input_stream_peek(). No fix
       available. The server does not use GLib's GIO input streams.
 
-  # gnupg2 - Present in UBI9 base image but not used by the server.
+  # gnupg2 - Present in UBI9 base image but not used by the server; it
+  # does not verify GPG signatures or process OpenPGP data at runtime
+  # (the pgedge RPM repo's GPG key is only checked by rpm/microdnf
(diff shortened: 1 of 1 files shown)


tsivaprasad (approved):
Looks good

PR #209, opened 2026-07-30
fix(security): trim unused packages from nla-web, refresh ignore list

moizpgedge:
## What this fixes

A security scan of the published web app image flagged 9 Critical,
63 High, 279 Medium, and 431 Low warnings (782 total).

Rebuilding the image from the current version, with no other changes,
already fixes most of the fixable ones, including the single Critical
warning.

## Going further: removed software the app doesn't use

Two pieces of included software accounted for most of what was left:

- A text editor that ships with a long history of unfixed security
  warnings, included in the image but never used or needed.
- A set of optional add-ons for the web server (mail handling, a
  scripting language, image processing, etc.). I checked the app's
  actual configuration file and confirmed none of these are turned on
  or used anywhere.

Removed both, along with a few other unused tools that came along with
them. Result: the image now includes about a third fewer installed
packages, and warnings dropped from 759 to 290, with zero Critical and
zero High remaining.

Two small pieces were deliberately left in place, since removing them
would require also removing core system tools the operating system
depends on to manage itself — a bigger risk for a smaller benefit.
These are documented as "checked, doesn't apply" instead.

## Tested, not just built

After trimming the image, I actually ran it and made real requests to
it: the app started cleanly, served the correct page, and the correct,
working files, exactly as before. I also confirmed a known, unrelated
quirk (it can't find a service name when run standalone outside its
normal setup) already existed before this change and isn't something
this PR introduced.

## Results

Scanning the newly built, trimmed image against the updated list of
"checked, doesn't apply" items shows zero warnings, at every severity
level.

## What this doesn't do

happens.
https://pgedge.atlassian.net/browse/PLAT-702

### DIFF
--- a/.trivy/nla-web.trivyignore.yaml
+++ b/.trivy/nla-web.trivyignore.yaml
@@ -1,314 +1,437 @@
 vulnerabilities:
-  # nginx - Runtime web server; serves static HTML/JS/CSS files only.
+  # nginx - Runtime web server; serves static HTML/JS/CSS files and
+  # proxies /mcp/v1 only. docker/nginx.conf uses a plain http block:
+  # no mail{}/stream{} block, no mp4/image-filter/xslt/perl module
+  # directives, and no OCSP configuration.
   - id: CVE-2025-23419
     statement: >-
       TLS session resumption vulnerability. Risk accepted; TLS termination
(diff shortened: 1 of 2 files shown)


tsivaprasad (approved):
Looks good.

PR #210, opened 2026-07-30
fix(security): refresh nla-cli Trivy ignore list

moizpgedge:
## What this fixes

A security scan of the published command-line tool image flagged 27
High, 32 Medium, and 10 Low warnings.

Rebuilding the tool from the current version, with no code changes,
already fixes every one of these — they were all caused by software
updates that had already happened but hadn't made it into a new build
of this image yet.

## What was left

A fresh scan of the rebuilt image shows only 10 minor warnings left:

- 9 are in small pieces of the base operating system that this tool
  doesn't actually use (no fix is available for these from the vendor
  yet, and none apply here).
- 1 is an advisory flagging an old, unmaintained piece of an encryption
  library as something people shouldn't use going forward. It's not a
  security bug with a fix — it's a "don't use this anymore" notice.
  Checked the tool's code directly: it only uses a different, unrelated
  part of that same library, never the flagged part.

Each of these is documented with a specific, checkable reason rather
than a generic one.

## Results

Scanning the newly built image against the updated list of "checked,
doesn't apply" items shows zero warnings, at every severity level.

## What this doesn't do

happens.
https://pgedge.atlassian.net/browse/PLAT-703

## Summary by CodeRabbit

* **Chores**
  * Updated vulnerability scan exceptions for several packages and components.
  * Added expiration dates, vulnerability details, and usage rationale to the updated entries.

### DIFF
--- a/.trivy/nla-cli.trivyignore.yaml
+++ b/.trivy/nla-cli.trivyignore.yaml
@@ -1,17 +1,56 @@
 vulnerabilities:
   # coreutils-single - Present in UBI9 micro base image.
-  - id: CVE-2025-5278
+  - id: CVE-2026-56391
     statement: >-
-      Heap buffer under-read in coreutils sort. No fix available. The
-      CLI container does not invoke sort with untrusted input.
+      GNU coreutils uniq out-of-bounds read with multibyte input via -w.
+      No fix available. The CLI does not invoke coreutils uniq.
+  - id: CVE-2026-56392
+    statement: >-
+      GNU coreutils unexpand heap buffer overflow via crafted -t tab stop
+      values. No fix available. The CLI does not invoke coreutils
+      unexpand.
 
-  # glibc - Core C library in UBI9 base image.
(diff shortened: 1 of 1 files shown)


.trivy/nla-cli.trivyignore.yaml
```diff
@@ -1,17 +1,56 @@
 vulnerabilities:
   # coreutils-single - Present in UBI9 micro base image.
-  - id: CVE-2025-5278
+  - id: CVE-2026-56391
```
tsivaprasad, line 3: For consistency with #209 (which added purls: scoping and expired_at: expiry to every ignore entry), please add the same to the entries here as well.

tsivaprasad (commented):
For consistency with #209 (which added purls: scoping and expired_at: expiry to every ignore entry), please add the same to the entries here as well.

tsivaprasad (approved):
Awesome...!
Looks good.

PR #213, opened 2026-08-04
fix: return tools/prompts/resources in a deterministic sorted order

### LINKED ISSUE #211: tools/list returns a nondeterministic order, defeating client and prompt caching
## Summary

`tools/list` returns tools in a different order on every call. The tool
registry stores tools in a Go map and `List()` iterates it directly, so
the resulting slice order is randomised by the runtime:

https://github.com/pgEdge/pgedge-postgres-mcp/blob/main/internal/tools/registry.go#L52-L58

```go
func (r *Registry) List() []mcp.Tool {
    tools := make([]mcp.Tool, 0, len(r.tools))
    for _, tool := range r.tools {
        tools = append(tools, tool.Definition)
    }
    return tools
}
```

Every `tools/list` path reaches this method.
`ContextAwareProvider.ListContext` returns either `p.baseRegistry.List()`
or a per-client `registry.List()`, and both are the same map-iterating
implementation, so the reshuffling affects authenticated and
unauthenticated callers alike.

## Why this matters

A reordered tool list changes the serialised tool definitions that a
client sends to its model on every request. Tool definitions normally sit
at the very front of the prompt, ahead of the system prompt and the
conversation, so reordering them invalidates the provider-side prompt
cache for the entire remainder of the request. Any agent that talks to
this server therefore pays full input-token price on turns that should
have been cache hits, and pays it on every turn rather than occasionally.

The order also defeats client-side caching and diffing of the tool list:
a client cannot cheaply tell whether the advertised set has actually
changed, because a byte comparison differs even when the set is
identical.

The MCP specification added explicit guidance on this in revision
2026-07-28: servers should return tools from `tools/list` in a
deterministic order, precisely to enable client-side caching and improve
prompt cache hit rates. The guidance is a `SHOULD` on the server, not a
protocol feature, so adopting it does not require the server to advertise
any newer protocol version. See
https://modelcontextprotocol.io/specification/2026-07-28/changelog
(Minor changes, item 3).

## Scope

The same pattern applies to the prompt and resource registries, and the
caching argument is identical for `prompts/list` and `resources/list`:

- `internal/prompts/registry.go:50-56`
- `internal/resources/registry.go:52-58`

`internal/resources/context_aware_registry.go:54` builds its static
built-in entries in a fixed literal order, but it should be checked for
whether any dynamically discovered entries are appended from a map.

## Suggested fix

Sort by the natural stable key before returning: tool name, prompt name,
and resource URI respectively. Sorting inside `List()` keeps the fix in
one place per registry and covers every caller, rather than sorting at
each `tools/list` handler.

Worth adding a test that calls `List()` repeatedly and asserts the order
is identical across calls; a single call cannot detect this, since any
one map iteration looks fine on its own.

moizpgedge:
## Summary

`tools/list`, `prompts/list`, and `resources/list` returned entries in a
different order on every call, because each registry built its response by
iterating a Go map directly. Tool, prompt, and resource definitions sit at
the front of the prompt sent to the model, so the reshuffling invalidated
the provider-side prompt cache on every single request rather than only
when the advertised set actually changed, and it defeated client-side
diffing of the list for the same reason.

Fixes #211.

## Root cause

```go
func (r *Registry) List() []mcp.Tool {
    tools := make([]mcp.Tool, 0, len(r.tools))
    for _, tool := range r.tools {
        tools = append(tools, tool.Definition)
    }
    return tools
}
```

`r.tools` is a `map[string]Tool`. Go deliberately randomizes map iteration
order, so this produced a different slice order on every call. The same
pattern existed in three other places:

- `internal/tools/registry.go`
- `internal/prompts/registry.go`
- `internal/resources/registry.go`
- `internal/resources/context_aware_registry.go` (merges one fixed
  built-in entry with a map of custom resources)

## The fix

`List()` in each of the four places now sorts before returning: by name
for tools/prompts, by URI for resources. A follow-up CodeRabbit review
pointed out that sorting by name/URI alone still leaves the relative order
of two entries undefined if they ever advertised the *same* name/URI under
different registration keys, since `sort.Slice` gives no ordering
guarantee between elements that compare equal. I checked every `Register()`
call site in this codebase (built-in tools, custom tools, custom
resources) and confirmed the registration key is always identical to the
advertised name/URI today, so that scenario can't currently occur — but
added the registration key as a secondary sort field anyway, since it's
free (the tiebreak never activates for any caller today) and makes `List()`
provably deterministic for any future caller too.

## Does this affect the CLI or web client?

No visible output changed. Both clients already defended themselves
against this independently:

- CLI: `/list tools`, `/list resources`, and `/list prompts`
  (`internal/chat/commands.go`) already built a locally sorted copy before
  printing.
- Web: the only visible list is the prompt picker
  (`web/src/components/PromptPopover.jsx`), which already does
  `[...prompts].sort(...)` before rendering. Tools/resources are never
  rendered as a list in the web client at all.

The bug only cost prompt-cache efficiency — full input-token price on every
turn instead of only when the tool set changed — never correctness of what
a user saw on screen.

## Testing

- Added `TestList_DeterministicOrder` and
  `TestList_DeterministicOrder_EqualNames`/`EqualURIs` to all four
  registries, asserting sorted order across 50 repeated calls. Confirmed
  each fails reliably against the unfixed code and passes against the fix.
- Reproduced the bug against the actual running server before fixing it:
  15 real `tools/list` calls over HTTP against the unfixed binary came
  back in a different rotation almost every time. After the fix, 20 calls
  in a row — including with custom tools/resources loaded from a
  definitions file, mixed in among the built-ins — all return
  byte-identical JSON.
- Checked the stdio transport too: 15 independent fresh process
  invocations (each gets its own random map seed at startup) all produced
  the identical order.
- `go build`/`go vet`/`gofmt` clean, `golangci-lint`: 0 issues, `go test
  -race` clean, `make test`: all 21 packages pass, `mkdocs build
  --strict` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * List results for tools and prompts are now consistently sorted by name.
  * Resource lists are now consistently sorted by URI.
  * Duplicate names or URIs receive stable ordering, preventing results from changing between requests.
  * Improves provider prompt caching and client-side list comparisons.

* **Documentation**
  * Added changelog details for the deterministic ordering improvements.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -161,6 +161,21 @@ and this project adheres to
   `httperror`, `llmtracing`, `logging`, `prompts`, `search` and `tsv`. All of
   them pass; they were simply never run.
 
+- `tools/list`, `prompts/list`, and `resources/list` now return their
+  entries in a stable, sorted order on every call. Each registry stored
+  its entries in a Go map and built the response by iterating it
+  directly, so the order was randomised by the runtime and changed from
+  one call to the next even when the underlying set was unchanged. Tool,
+  prompt, and resource definitions sit at the front of the prompt sent to
+  the model, so the reshuffling invalidated the provider-side prompt
+  cache on every request rather than only when the set actually changed,
+  and it defeated client-side diffing of the advertised list for the
+  same reason. `Registry.List()` in each of `internal/tools`,
+  `internal/prompts`, and `internal/resources` now sorts by name (tools,
+  prompts) or URI (resources) before returning; `ContextAwareRegistry.List()`
+  in `internal/resources`, which merges a fixed built-in entry with a map
+  of custom resources, sorts its combined result the same way. Fixes #211.
+
 ### Added
 
 - A `make vulncheck` target runs `govulncheck` over the module, using
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -174,7 +174,15 @@ and this project adheres to
   `internal/prompts`, and `internal/resources` now sorts by name (tools,
   prompts) or URI (resources) before returning; `ContextAwareRegistry.List()`
   in `internal/resources`, which merges a fixed built-in entry with a map
-  of custom resources, sorts its combined result the same way. Fixes #211.
+  of custom resources, sorts its combined result the same way. Every
+  registration is keyed by its own name or URI in this codebase today,
+  so the sort has no ties to break in practice, but the comparison also
+  falls back to the registration key, which is unique by construction:
+  `sort.Slice` gives no ordering guarantee between two entries that
+  compare equal, so a future entry whose Definition happened to
+  advertise the same name or URI as another under a different key would
+  otherwise be exactly as nondeterministic as before this fix. Fixes
+  #211.
 
 ### Added
 
--- a/internal/prompts/registry.go
+++ b/internal/prompts/registry.go
@@ -46,12 +46,17 @@ func (r *Registry) Get(name string) (Prompt, bool) {
 	return prompt, exists
 }
 
-// List returns all registered prompt definitions
+// List returns all registered prompt definitions, sorted by name for a
+// deterministic order across calls; see the tool registry's List for why
+// this matters for prompt caching.
 func (r *Registry) List() []mcp.Prompt {
 	prompts := make([]mcp.Prompt, 0, len(r.prompts))
 	for _, prompt := range r.prompts {
 		prompts = append(prompts, prompt.Definition)
 	}
+	sort.Slice(prompts, func(i, j int) bool {
+		return prompts[i].Name < prompts[j].Name
+	})
 	return prompts
 }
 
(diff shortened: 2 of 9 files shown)


dpage (approved):
Reviewed and verified locally: `gofmt`, `go vet` and the tools/prompts/resources package tests are all clean. I checked the new tests are genuinely load-bearing by reverting `tools.Registry.List()` to the original map iteration (both new tests fail) and separately by swapping the tiebreak from the registration key to `Description` (the equal-names test fails), so they really do pin the behaviour they claim to.

I also traced every consumer of the four `List()` implementations: `internal/mcp/server.go`, `internal/mcp/http_server.go`, `ContextAwareProvider.List`/`ListContext`, and the `read_resource` tool's `list=true` path all funnel through them, so there is no remaining unsorted path. Both CodeRabbit findings are addressed and resolved, and I agree with declining the validation half of the first one.

Two non-blocking observations for a future tidy-up rather than this PR: the keyed-sort block is repeated near-verbatim in four places and would collapse into a single generic helper, and `List()` still reads the map without a lock, which is pre-existing and unreachable today given registries are fully populated before being published, but worth keeping in mind.

PR #214, opened 2026-08-04
fix: negotiate protocolVersion instead of echoing the client's request

### LINKED ISSUE #212: initialize echoes the client's requested protocolVersion instead of one the server supports
## Summary

`handleInitialize` returns whatever `protocolVersion` the client asked
for, rather than a version the server actually implements:

https://github.com/pgEdge/pgedge-postgres-mcp/blob/main/internal/mcp/server.go#L190-L194

```go
// Accept the client's protocol version for compatibility
protocolVersion := params.ProtocolVersion
if protocolVersion == "" {
    protocolVersion = ProtocolVersion
}
```

The server implements `2024-11-05` (`internal/mcp/server.go:25`), but a
client requesting any other revision is told that revision was accepted.

## Why this is wrong

Version negotiation in MCP is the server's half of the handshake. The
client proposes a version; the server replies with the version it will
actually speak, which may be older than the one proposed. The client then
decides whether it can work with that answer, and disconnects if it
cannot. Echoing the request removes the client's ability to make that
decision, because the answer carries no information.

The practical consequence is a failure that surfaces late and in the
wrong place. A client asking for `2025-11-25` is told `2025-11-25` and
proceeds to use features from the four revisions we do not implement.
Instead of a clean negotiation failure at connect time, it gets a
method-not-found or a missing field partway through a real request, and
the error points at the feature rather than at the version mismatch that
caused it.

The comment calls this a compatibility measure, and in practice it is why
newer clients work against this server at all today. Replying honestly
with `2024-11-05` is the compatible behaviour, though: that is exactly
the case negotiation is designed to handle, and a client that supports a
newer revision is expected to fall back.

## Suggested fix

Reply with the highest version this server supports. A minimal change is
to return `ProtocolVersion` unconditionally, which is correct as long as
the server implements exactly one revision.

If we would rather be explicit, a small set of supported versions plus
"pick the newest one at or below what the client asked for, otherwise the
oldest we support" generalises cleanly to the point where we implement
more than one revision. That structure is also what a future
`server/discover` implementation would advertise, so it is not wasted
work.

Either way this only affects the value returned in `InitializeResult`;
no request handling changes.

## Related

This does not require adopting a newer protocol revision, and is worth
fixing independently of that decision. Noting for context that the
server is currently four revisions behind (`2025-03-26`, `2025-06-18`,
`2025-11-25`, `2026-07-28`), and that the incoming
`MCP-Protocol-Version` HTTP header introduced in `2025-06-18` is not
read anywhere either.

moizpgedge:
## Summary

`initialize` returned whatever `protocolVersion` a client asked for, rather
than a version the server actually implements. Version negotiation is the
server's half of the MCP handshake: the client proposes a revision, the
server replies with the revision it will actually speak, and the client
checks that answer against what it supports. Echoing the request answers
with no information at all, so a client asking for a revision this server
does not implement was told that revision was accepted, and found out
otherwise later — against a missing capability — rather than at the
version check meant to catch exactly this case.

Fixes #212.

## Root cause

```go
// Accept the client's protocol version for compatibility
protocolVersion := params.ProtocolVersion
if protocolVersion == "" {
    protocolVersion = ProtocolVersion
}
```

Reproduced against the actual running server before fixing anything:
stdio echoed a bogus `"2099-01-01"` request back verbatim. The same
request over HTTP already correctly returned `"2024-11-05"` — that
asymmetry turned out to be because `handleInitializeHTTP` never parsed
`req.Params` at all; it just always returned the hardcoded constant,
which only happened to be harmless because this server currently
implements exactly one revision.

## The fix

Added `SupportedProtocolVersions` (oldest to newest) and
`NegotiateProtocolVersion()`: returns the newest supported revision at or
below the client's request, or the oldest supported revision if the
request predates everything this server implements. Wired both
`handleInitialize` (stdio) and `handleInitializeHTTP` through it — the
latter now actually reads the client's request for the first time.

Verified the HTTP-side change has real, if not yet observable, value:
temporarily added a second fake supported version and confirmed
negotiation correctly returns the older one for an old request, where the
previous hardcoded-constant behavior would have kept wrongly returning the
newest regardless of what the client asked for.

## A second bug found while re-verifying

Re-checking every angle turned up a real bug in the fix itself:
`handleInitializeHTTP`'s new params-parsing silently swallowed unmarshal
errors (e.g. `protocolVersion` sent as a number instead of a string),
returning `200` with the default version instead of a proper JSON-RPC
error — inconsistent with stdio's `handleInitialize` and every other
`*HTTP` handler in this file, which all return `-32602` on malformed
params. Fixed to match, and verified by reverting the fix and confirming
the new regression test fails.

Both transports now handle three scenarios identically:

| Request | Result |
|---|---|
| Malformed type (`protocolVersion: 12345`) | `-32602 Invalid params` |
| Unsupported future version (`"2099-01-01"`) | Negotiated down to `2024-11-05` |
| Omitted params entirely | Default (`2024-11-05`) succeeds |

## Does this affect the CLI or web client?

No. Both request `2024-11-05` exactly and get back exactly what they
asked for either way.

## Not done here

Per the issue, explicitly out of scope: reading the `MCP-Protocol-Version`
HTTP header (2025-06-18+), or adopting a newer protocol revision.

## Testing

- Reproduced the original bug live against the real running binary
  (stdio vs. HTTP asymmetry) before writing any fix.
- Added `TestNegotiateProtocolVersion` (table-driven),
  `TestNegotiateProtocolVersion_NeverEchoesUnsupported`,
  `TestSupportedProtocolVersions_NewestIsProtocolVersion` (guards drift
  between the list and the constant), and dedicated regression tests per
  transport for both the negotiation fix and the malformed-params fix.
- Proved every regression test actually catches its own bug: reverted the
  relevant file, confirmed the test fails, restored, confirmed it passes.
- Live-verified all three scenarios above against the real binary on both
  transports after the fix.
- `go build`/`go vet`/`gofmt` clean, `golangci-lint`: 0 issues, `go test
  -race`: all 79 tests in `internal/mcp` pass, `make test`: all 21
  packages pass, `mkdocs build --strict` clean.

## Summary by CodeRabbit

- **New Features**
  - Added MCP protocol version negotiation during initialization for both HTTP and standard input/output connections.
  - Unsupported or omitted client versions now fall back to a compatible server-supported version.

- **Bug Fixes**
  - Malformed initialization parameters now return a clear JSON-RPC `-32602 Invalid params` error.
  - Omitted initialization parameters are accepted using server defaults.

- **Documentation**
  - Documented protocol negotiation, fallback behavior, and initialization validation.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -181,11 +181,15 @@ and this project adheres to
   and web client are unaffected, since both already request
   `2024-11-05` exactly.
 
-  `initialize` over HTTP also now rejects a `protocolVersion` of the
-  wrong JSON type with a proper `-32602 Invalid params` error, matching
-  every other HTTP handler in this file and the stdio transport's
-  existing behaviour, instead of silently falling back to the server's
-  default as if the field had been omitted.
+  `initialize` over HTTP also now rejects malformed parameters with a
+  proper `-32602 Invalid params` error, matching every other HTTP handler
+  and the stdio transport's existing behaviour. The HTTP handler
+  previously never read the client's parameters at all, so it accepted
+  anything; it now parses them, and a field of the wrong JSON type, such
+  as a numeric `protocolVersion` or a `clientInfo` that is not an object,
+  fails the handshake instead of silently falling back to the server's
+  default. Omitting the parameters entirely still succeeds and yields
+  that default.
 
 ### Added
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -161,6 +161,32 @@ and this project adheres to
   `httperror`, `llmtracing`, `logging`, `prompts`, `search` and `tsv`. All of
   them pass; they were simply never run.
 
+- `initialize` no longer echoes back whatever `protocolVersion` a client
+  requests. Version negotiation is the server's half of the MCP
+  handshake: the client proposes a revision, and the server is supposed
+  to reply with the revision it will actually speak, which the client
+  then checks against what it supports. Echoing the request answers
+  with no information at all, so a client asking for a revision this
+  server does not implement was told that revision was accepted, and
+  found out otherwise later, against a missing capability, rather than
+  at the version check meant to catch exactly this case. `initialize`
+  (stdio) and `initialize` over HTTP now both call a shared
+  `NegotiateProtocolVersion`, which returns the newest revision this
+  server supports at or below the client's request, or the oldest
+  supported revision if the request predates everything the server
+  implements. This server currently implements one revision,
+  `2024-11-05`, so today every negotiation converges on that value
+  regardless of what a client asked for; the function is structured to
+  extend cleanly if a second revision is added later. The bundled CLI
+  and web client are unaffected, since both already request
+  `2024-11-05` exactly.
+
+  `initialize` over HTTP also now rejects a `protocolVersion` of the
+  wrong JSON type with a proper `-32602 Invalid params` error, matching
+  every other HTTP handler in this file and the stdio transport's
+  existing behaviour, instead of silently falling back to the server's
+  default as if the field had been omitted.
+
 ### Added
 
 - A `make vulncheck` target runs `govulncheck` over the module, using
--- a/docs/developers/mcp-protocol.md
+++ b/docs/developers/mcp-protocol.md
@@ -6,6 +6,15 @@ This document describes how the Natural Language Agent implements the Model Cont
 
 This server implements **MCP version `2024-11-05`**.
 
+The server negotiates the version during the `initialize` handshake; it
+never simply echoes the version a client requests. A client that asks for
+a revision the server does not implement receives `2024-11-05` in the
+`initialize` response, and it should then decide whether to continue or
+disconnect. A client that omits `protocolVersion` also receives
+`2024-11-05`. An `initialize` request whose parameters are malformed, such
+as a `protocolVersion` that is not a string, fails with a `-32602 Invalid
+params` error on both transports.
+
 ## Transport Modes
 
 The server supports two transport modes:
--- a/internal/mcp/http_server.go
+++ b/internal/mcp/http_server.go
@@ -382,6 +382,15 @@ func (s *Server) handleRequestHTTP(ctx context.Context, req JSONRPCRequest) JSON
 // HTTP-specific handlers that return responses instead of sending them
 
 func (s *Server) handleInitializeHTTP(req JSONRPCRequest) JSONRPCResponse {
+	var params InitializeParams
+	paramsJSON, err := json.Marshal(req.Params)
+	if err != nil {
+		return createErrorResponse(req.ID, -32602, "Invalid params", err.Error())
+	}
+	if err := json.Unmarshal(paramsJSON, &params); err != nil {
+		return createErrorResponse(req.ID, -32602, "Invalid params", err.Error())
+	}
+
 	capabilities := map[string]interface{}{
 		"tools": map[string]interface{}{},
 	}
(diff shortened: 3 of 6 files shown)


dpage (approved):
Approving after a review-driven cleanup, pushed as `40319ec` on top of the author's work.

The fix itself is right, and rather more valuable than the PR claims. Revision `2026-07-28`, released a week or so ago, removes the `initialize` handshake altogether: a client now declares its version in per-request `_meta` and the `MCP-Protocol-Version` header, servers must implement `server/discover`, and mismatches return `UnsupportedProtocolVersionError` (`-32022`). Dual-era clients therefore detect a legacy server by probing and falling back to `initialize`, proposing their newest *legacy* revision, which will be `2025-11-25`. Under the old echo behaviour this server would have replied `"2025-11-25"` and the client would then have used four revisions' worth of features this server has never implemented. That fallback path is reachable today: both transports return `-32601` for an unknown method (`internal/mcp/server.go:174`, `internal/mcp/http_server.go:378`), which is exactly the non-modern error a dual-era client reads as 'legacy server'. So this bug was about to start firing routinely rather than theoretically.

What the cleanup commit changes, none of it behavioural:

- `SupportedProtocolVersions` described itself as every revision the server implements, which invites appending `2026-07-28` alongside a `ProtocolVersion` bump; `NegotiateProtocolVersion` would then answer a legacy handshake with a revision that has no handshake. It is now documented as handshake-era only, guarded by `TestSupportedProtocolVersions_AreAllLegacy`, and unexported whilst renaming, since nothing outside the package reads it and its ordering is load-bearing.
- The doc comment claimed to implement the specification's negotiation rule. It does not: the spec, worded identically across `2024-11-05`, `2025-06-18` and `2025-11-25`, says reply with the requested revision if supported, otherwise a supported one, which SHOULD be the server's latest. 'Newest at or below the request' is a deliberately more conservative reading, which I think is the better choice; the comment now states the deviation and its reasoning rather than claiming conformance.
- The changelog said the HTTP handler rejects a `protocolVersion` of the wrong type. It rejects any malformed params, since the whole struct goes through one `Unmarshal` and the handler previously never read params at all. Reworded, and `docs/developers/mcp-protocol.md` now documents the negotiation behaviour.

I verified the original work rather than taking it on trust: reverting the stdio handler to the echo behaviour makes `TestHandleInitialize_Stdio_NegotiatesProtocolVersion` fail, and restoring the swallowed unmarshal error makes `TestHandleInitializeHTTP_MalformedProtocolVersion` fail. Worth noting that `TestHandleInitializeHTTP_NegotiatesProtocolVersion` cannot fail against the pre-fix code, since that hardcoded the same constant; it is a valid guard rather than a red-first regression test.

Note on CI: GitHub Actions has scheduled no workflow run for this repository since the `main` push at 12:07, so the five Actions checks did not run against `40319ec`, and a close/reopen did not shake them loose. They were green on the parent commit `ba2d693`. I verified locally instead: `gofmt` and `go vet ./...` clean, `golangci-lint` 0 issues, `go test -race` over the CI package set clean, `make test` all 21 packages passing (exit 0), and `mkdocs build --strict` clean. Codacy passed on the new head.

Follow-ups worth their own issues rather than this PR: adopting `2026-07-28` is substantial (mandatory `server/discover`, sessions and `Mcp-Session-Id` removed, `ping` and `logging/setLevel` gone, `resultType` on all results, `ttlMs`/`cacheScope` on list results, resource-not-found moving from `-32002` to `-32602`), and it is worth confirming that real dual-era clients fall back correctly over HTTP, where the spec's mechanic expects a `4xx` and this server would return 200 with a `-32601` body.

PR #215, opened 2026-08-04
feat: add dual-era support for the MCP 2026-07-28 spec revision

moizpgedge:
## Summary

Brings the server up to the current MCP spec revision (`2026-07-28`) while leaving every existing `2024-11-05` client working exactly as before. The two revisions are served side by side on the same endpoint, per the versioning spec's own dual-era guidance: a request is handled as modern only if it opts in via `_meta.io.modelcontextprotocol/protocolVersion`; everything else takes the legacy path, unchanged.

Full decision record — what's implemented, what was deliberately left out, and why — lives in [`docs/developers/mcp-spec-compliance.md`](docs/developers/mcp-spec-compliance.md).

## What's new

- **`server/discover`**, required by the new spec for every server, on both stdio and HTTP.
- **Per-request `_meta` negotiation** — `protocolVersion`, `clientCapabilities`, `clientInfo`, `logLevel` — replacing the old connection-scoped `initialize` handshake for any client that opts in.
- **Required Streamable HTTP headers** (`MCP-Protocol-Version`, `Mcp-Method`, `Mcp-Name`), including the Base64 sentinel encoding for header values that aren't safe as plain ASCII.
- **`resultType` / `ttlMs` / `cacheScope`** on every modern result, per the spec's caching model.
- **Three new JSON-RPC error codes**: `-32020 HeaderMismatch`, `-32021 MissingRequiredClientCapability`, `-32022 UnsupportedProtocolVersion`.

## Bugs found along the way

Two independent, pre-existing bugs turned up while implementing this and are fixed here too:

1. **`resources/read` on an unknown URI returned a disguised success**, not an error — the response was a normal JSON-RPC result whose content happened to say "not found" in text. It's now a real JSON-RPC error (`-32002` legacy / `-32602` modern, per the spec's own renumbering). A second, compounding bug in the same path required a live database connection just to report that a resource didn't exist; the existence check now happens before any database access.
2. **A malformed `Mcp-Name` header could panic the request handler** instead of being rejected with a clean `400`. The Base64 sentinel's prefix and suffix share a character, so a short malformed value could satisfy both boundary checks at once and produce a negative-length slice. Caught by `recoveryMiddleware` rather than crashing the server, but it turned a client mistake into a `500` with a logged stack trace instead of the correct `400 HeaderMismatch`.

## What was deliberately not adopted

Recorded in full in the compliance doc, briefly: subscriptions/listen (nothing in this server changes at runtime to notify about), Multi Round-Trip Requests and the client-side features it replaces (never implemented, and the spec deprecates them anyway), `x-mcp-header` (explicitly optional for servers), OAuth-related changes (this server uses bearer tokens), icons (cosmetic), and the intermediate legacy revisions between `2024-11-05` and `2026-07-28` (all still handshake-based; a client needing more can use the stateless model instead).

## Testing

- New unit tests (`internal/mcp/modern_test.go`, additions to `internal/mcp/server_test.go`) and an 8-scenario HTTP integration suite (`test/mcp_modern_protocol_test.go`) covering `server/discover`, header validation, version negotiation, and era-appropriate error codes.
- Every fix was verified live against the running server binary (both transports) before and after, including a mutation-test pass: reverting each fix and confirming its regression test actually fails, then restoring it.
- `go build`, `go vet`, `gofmt -l`, `golangci-lint run` (0 issues), `go test ./...` and `go test -race ./...` all pass.

## Summary by CodeRabbit

- **New Features**
  - Added modern MCP protocol negotiation alongside legacy compatibility.
  - Added stateless requests, server discovery, response metadata, and caching details.
  - Added validation for protocol headers and request metadata.
  - Expanded API documentation for supported MCP methods and response statuses.

- **Bug Fixes**
  - Unknown resources now return proper protocol errors.
  - Improved handling of malformed headers, unsupported versions, missing capabilities, and unknown methods.
  - Preserved legacy request behavior while improving modern request handling.

- **Documentation**
  - Expanded MCP protocol and specification compliance guidance.
  - Added compliance documentation to developer navigation.


internal/mcp/http_server.go
```diff
@@ -319,11 +319,42 @@ func (s *Server) handleHTTPRequest(w http.ResponseWriter, r *http.Request) {
...
+	if !preflightRejected {
+		response = s.handleRequestHTTP(ctx, req)
+		if isModern && response.Error == nil {
+			response.Result = wrapModernResultForMethod(req.Method, response.Result)
+		}
+	}
+
+	statusCode := http.StatusOK
+	if isModern && preflightRejected {
+		statusCode = http.StatusBadRequest
```
dpage, line 364: **Must fix: a modern `-32601` needs HTTP `404`, not `200`.**

`statusCode` only leaves `200` when `preflightRejected` is true, so every
`-32601` produced inside `handleRequestHTTP` goes out as `200`: the new
modern-`initialize` rejection at `:391-392`, the unknown-method default at
`:420`, and the "Prompts not supported" / "Database management not supported"
branches at `:579`, `:594`, `:634` and `:656`.

The Streamable HTTP binding for this revision is explicit: "If the server does
not implement the requested RPC method, it MUST respond with `404 Not Found`
and a JSON-RPC error with code `-32601` (Method not found). The JSON-RPC error
body distinguishes this case from a `404` returned by a legacy HTTP+SSE server
that does not host the modern MCP endpoint." That status is therefore
load-bearing in the client-side era-detection and fallback flow described in
the same section, so a modern client probing this server currently gets an
ambiguous signal.

Small fix: after dispatch, when `isModern && response.Error != nil &&
response.Error.Code == -32601`, set `statusCode = http.StatusNotFound`. Note
that this also inverts the assertion in
`TestModernProtocol_Initialize_RejectedAsMethodNotFound`, which currently
asserts `200` with a comment explaining why, so the test enshrines the status
that needs changing. Worth adding a case for a genuinely unknown method under
modern `_meta` at the same time, since nothing covers that path today.

internal/mcp/server.go
```diff
@@ -295,7 +352,7 @@ func (s *Server) handleToolsList(req JSONRPCRequest) {
 		"tools": tools,
 	}
 
-	sendResponse(req.ID, result)
+	sendResponseFor(req, result, true)
```
dpage, line 355: Whilst you were moving handlers onto `sendResponseFor`, `handlePing` at
`:342-343` was left on the bare `sendResponse`, so a modern stdio ping returns
`{}` with no `resultType`. The base specification is unconditional: "The
`result` MUST include a `resultType` field to indicate the type of the result."

The same ping over HTTP does get wrapped, because `handleHTTPRequest` wraps
centrally by method name, so the two transports presently disagree about an
identical request. One-liner:
`sendResponseFor(req, map[string]interface{}{}, false)`.

test/mcp_modern_protocol_test.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	params := map[string]interface{}{
+		"protocolVersion": "2024-11-05",
+		"_meta": map[string]interface{}{
+			"io.modelcontextprotocol/protocolVersion":    "2026-07-28",
+			"io.modelcontextprotocol/clientCapabilities": map[string]interface{}{},
+		},
+	}
+	status, resp := sendModernHTTPRequest(t, server, "initialize", params, headers)
+	if status != http.StatusOK {
```
dpage: This assertion locks in the status that finding 1 in my review needs changed:
the specification requires `404 Not Found` alongside `-32601` for a method the
server does not implement, and it is the pairing of that status with a
recognisable modern JSON-RPC error body that lets a client tell a modern server
apart from a legacy HTTP+SSE server which simply does not host this endpoint.
The distinction between a preflight rejection and a handler-level one is a real
and useful one, it just does not map onto `200` here.

dpage (commented):
I've read the full diff along with the surrounding code, and I checked every
specification claim in the PR and in `mcp-spec-compliance.md` against the live
normative pages for `2026-07-28` rather than taking them on trust.

This is a genuinely good piece of work. The dual-era split, with the body
`_meta` field as the era discriminator and the legacy path left alone, is
exactly what the versioning specification's own guidance contemplates ("A
dual-era server MAY serve both eras concurrently on the same endpoint or
process"), and the two incidental bug fixes are real bugs with proper
regression tests behind them. The quotes in the compliance document all hold
up, including the `-32020`/`-32021`/`-32022` allocations, the `-32002` to
`-32602` renumbering, the `DiscoverResult` shape with `serverInfo` living under
`_meta` rather than at the top level, and the `UnsupportedProtocolVersionError`
data shape. The decision record itself is the sort of thing I wish more PRs
left behind, since it means nobody has to re-derive this reasoning in a year's
time.

There is one real compliance defect, which unluckily sits in the code the last
commit added, plus a handful of smaller things.

## Must fix

**1. A modern request that ends in `-32601` must be answered with HTTP `404`,
not `200`.** See the inline comment on `internal/mcp/http_server.go`. The
Streamable HTTP binding for this revision is explicit that an unimplemented RPC
method **MUST** get `404 Not Found` alongside the `-32601` body, and it says
why: the JSON-RPC error body is what distinguishes this case from the `404` a
legacy HTTP+SSE server returns when it does not host the modern endpoint. That
makes the status load-bearing in the client-side era-detection and fallback
flow described in the same section, so this is not merely pedantic. It affects
the new modern-`initialize` rejection, the unknown-method default, and the
"Prompts not supported" and "Database management not supported" branches.

## Should fix

**2. A modern `ping` over stdio returns a result with no `resultType`.**
`internal/mcp/server.go:342-343`. The base specification is unconditional here:
"The `result` **MUST** include a `resultType` field to indicate the type of the
result." Every other stdio handler was moved onto `sendResponseFor`, but
`handlePing` still calls the bare `sendResponse`, so a modern stdio ping yields
`{}`. The same ping over HTTP *is* wrapped, because `handleHTTPRequest` wraps
centrally by method name, so the two transports currently disagree about an
identical request. The fix is a one-liner:
`sendResponseFor(req, map[string]interface{}{}, false)`.

**3. Era detection ignores the `MCP-Protocol-Version` header, so a half-modern
HTTP request is silently served as legacy.** `internal/mcp/http_server.go:335`.
Deciding era on the body `_meta` is the right primary signal, but the transport
specification makes that header mandatory on every modern POST and requires its
value to match the body `_meta` value. A client that sends
`MCP-Protocol-Version: 2026-07-28` whilst omitting `_meta`, or misspelling one
of those rather long reserved keys, currently receives a legacy `200` carrying
none of the modern fields, whereas the specification wants it rejected: `-32602`
with `400` for the missing required `_meta` field, or `-32020` for header and
body disagreeing. Treating a `MCP-Protocol-Version` header that names a modern
revision as also marking the request modern would close the trap without
touching any genuinely legacy client, since nothing before `2025-06-18` sends
that header at all. Either way it deserves a line in the compliance document,
which currently presents body `_meta` as the whole story.

## Suggestions

**4. `server/discover` is missing from the OpenAPI method enum.**
`internal/openapi/openapi.go:670-678`, and the method list in the endpoint
description at `:172`. The API reference and the API browser are generated from
this, so both now understate what the server accepts.

**5. `read_resource` no longer tells the model which URI failed.**
`internal/tools/read_resource.go:112-114` now surfaces "Error reading resource:
resource not found", where the old disguised-success path at least said
"Resource not found: pg://whatever". Given this tool exists specifically for
LLM clients, including the URI is worth the two extra words.

**6. `isModernRequest` re-marshals `params` on every call, and it is called up
to three times per request.** Once in the dispatcher, again in
`resourceNotFoundCode` (`internal/mcp/modern.go:117`) or the `initialize`
branch, and once more in `sendResponseFor` (`internal/mcp/server.go:584`). For a
`tools/call` carrying a large SQL string that is three full marshal and
unmarshal round trips to re-answer a question already settled at the top of
dispatch. Threading the `(meta, isModern)` pair through would tidy this up, and
would probably also take a bite out of the five medium complexity issues Codacy
flagged.

**7. `TTLPermanent` is named for permanence but is 24 hours.**
`internal/mcp/types.go:1335`. The comment explains the reasoning perfectly
well, but `TTLOneDay` or `TTLMaxCacheable` would not need the explanation.

## Notes

**8. The remaining open CodeRabbit thread is wrong, and the code is right.** It
asks for the `=?base64?` sentinel prefix to be matched case-insensitively,
citing SEP-2243. Both that SEP and the current transport page say the opposite:
"The prefix `=?base64?` and suffix `?=` indicate that the value is
Base64-encoded. These markers are case-sensitive and **MUST** appear exactly as
shown (lowercase)." The SEP's conformance table is blunter still, listing
"Non-lowercase prefix: `=?BASE64?SGVsbG8=?=`" against the expected behaviour
"Server treats as literal value, not Base64", which is precisely what
`decodeHeaderValue` does today. I'd reply on that thread with the quotes and
resolve it rather than change the code.

**9. `Origin` header validation is absent, and unrecorded.** The transport
specification's security section requires servers to validate `Origin` on all
incoming connections to prevent DNS rebinding, answering `403` when it is
present and invalid, and I can find no `Origin` handling anywhere in
`internal/mcp`. This predates the PR and sits outside what it set out to do, so
I'm not asking for it here; but the whole value of `mcp-spec-compliance.md` is
that nobody has to re-derive this, and the "deliberately not adopted" list
doesn't currently mention it. Worth either fixing separately or recording
there.

## Summary

Correctness is sound apart from the `-32601` status and the stdio `ping` gap;
no new security exposure that I can see, since header values are only reflected
back to the caller inside JSON and validation sits behind authentication; the
test coverage is unusually thorough, including the mutation-test pass and the
double-wrap corruption check. I'd like to see finding 1 land before this
merges, and finding 2 is cheap enough to come along with it. Everything else is
at your discretion.

internal/mcp/modern.go
```diff
@@ -0,0 +1,359 @@
...
+	// character, which would make the slice below negative-length.
+	if len(v) < len(prefix)+len(suffix) || !strings.HasPrefix(v, prefix) || !strings.HasSuffix(v, suffix) {
+		return v, true
+	}
+	encoded := v[len(prefix) : len(v)-len(suffix)]
+	decoded, err := base64.StdEncoding.DecodeString(encoded)
+	if err != nil {
+		return "", false
+	}
+	return string(decoded), true
```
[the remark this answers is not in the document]
dpage: I don't think this one holds, and I've gone back to the normative text rather
than relying on the SEP summary. The specification requires the opposite of
what's suggested here: the sentinel markers are case-sensitive and lowercase
only, so `decodeHeaderValue` accepting just `=?base64?` is correct as written.

From [Streamable HTTP, Value
Encoding](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#value-encoding)
in the `2026-07-28` revision:

> The prefix `=?base64?` and suffix `?=` indicate that the value is
> Base64-encoded. These markers are case-sensitive and **MUST** appear exactly
> as shown (lowercase).

SEP-2243 itself, which this comment cites, carries the identical sentence in
its "Base64 Encoding for Unsafe Values" section, and its conformance test cases
are blunter still. Under "Server Validation Edge Cases", "Base64 Decoding":

> | Non-lowercase prefix | `=?BASE64?SGVsbG8=?=` | Server treats as literal
> value, not Base64 |

That is precisely the current behaviour: `=?BASE64?...?=` falls through the
`strings.HasPrefix` check and is returned unchanged as a literal header value,
which then fails the body comparison and yields `-32020 HeaderMismatch`. A
conforming client would never emit an upper-case prefix in the first place, and
must in fact Base64-encode any plain-ASCII value that merely *looks* like the
sentinel, precisely so that the literal and encoded cases stay unambiguous.

Matching the prefix case-insensitively would actively break that: a legitimate
literal header value of `=?BASE64?SGVsbG8=?=` would be silently decoded instead
of compared as-is, which is the ambiguity the case-sensitivity rule exists to
prevent.

So no change here, and no upper-case or mixed-case regression cases: accepting
them would be the bug. Resolving this thread.

internal/mcp/http_server.go
dpage: Fixed in 92ddcadf, and more thoroughly than I asked for: rather than special-casing the modern-`initialize` rejection, the status is now chosen by a `switch` on the response error code in `handleRequestHTTP`, so the unknown-method default and the "Prompts not supported" and "Database management not supported" branches are covered too. The last two had no coverage at all before, because the real binary always registers both providers, and they now have direct regression tests. Resolving.

internal/mcp/server.go
dpage: Fixed in 92ddcadf: `handlePing` is on `sendResponseFor` now, so a modern stdio ping carries `resultType` and the two transports agree about an identical request. `TestModernProtocol_Ping_HasResultType` covers it. Resolving.

test/mcp_modern_protocol_test.go
dpage: Addressed in 92ddcadf alongside the fix itself: the assertion is `StatusNotFound` now, and `TestModernProtocol_UnknownMethod_RejectedWith404` was added beside it. Resolving.

internal/openapi/openapi.go
```diff
@@ -169,17 +216,30 @@ func buildMCPPath() M {
...
+			"description": "Accepts a JSON-RPC 2.0 request implementing the Model Context Protocol. Supported methods are initialize, server/discover, ping, tools/list, tools/call, resources/list, resources/read, prompts/list, prompts/get, pgedge/listDatabases, and pgedge/selectDatabase. Bearer token authentication is required when auth is enabled. Two protocol revisions are served on this same endpoint: a request whose body carries _meta.io.modelcontextprotocol/protocolVersion (or, on HTTP, a present MCP-Protocol-Version header) is served under the current, stateless revision (2026-07-28), which requires the MCP-Protocol-Version, Mcp-Method, and (for tools/call, resources/read, prompts/get) Mcp-Name headers described below; any other request is served under the original, handshake-based revision (2024-11-05), which sends none of those headers and negotiates via initialize instead. initialize itself exists only under that original revision: it is not a method the current revision recognizes.",
```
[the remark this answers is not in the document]
dpage: Agreed, and fixed in c6a896c. The transport is explicit that "the client **MUST** include an `Accept` header listing both `application/json` and `text/event-stream` as supported content types", so it is genuinely part of the modern request contract and belongs in the specification.

One qualification, which the new `AcceptHeader` parameter states rather than glosses over: the requirement binds the client, not the server, and there is no matching server-side MUST to reject a request that omits it. This server always answers with `application/json`, because it implements no streaming method, and it does not inspect `Accept` at all, so a request without the header is not rejected. Documenting it as simply "required" would have advertised a rejection that never happens, so the description says what the transport asks of a conforming client and what this server actually does with it. The same reasoning is recorded under "Known gaps" in the compliance document, since enforcement only becomes meaningful alongside a streaming response.

internal/openapi/openapi.go
```diff
@@ -169,17 +216,30 @@ func buildMCPPath() M {
...
 			"responses": mergeResponses(M{
 				"200": M{
-					"description": "JSON-RPC 2.0 response.",
+					"description": "JSON-RPC 2.0 response. This includes a JSON-RPC-level error (for example, a tool's own argument validation failure) that is not one of the two rejections documented below.",
+					"content":     jsonContent(ref("JSONRPCResponse")),
+				},
+				"400": M{
+					"description": "The request identified itself as the current (2026-07-28) revision -- via body _meta or the MCP-Protocol-Version header -- and failed validation before dispatch: a required header is missing or does not match the request body (HeaderMismatch, -32020), a required _meta field is missing (-32602), or the requested protocol version is not supported (UnsupportedProtocolVersion, -32022).",
+					"content":     jsonContent(ref("JSONRPCResponse")),
+				},
```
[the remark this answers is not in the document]
dpage: I do not think this one holds, and I have gone back to the normative text rather than the summary.

The specification is certainly clear about the mapping: "If processing a request requires a capability the client did not include in `io.modelcontextprotocol/clientCapabilities`, the server **MUST** return a `MissingRequiredClientCapabilityError` (`-32021`) whose `data.requiredCapabilities` lists the missing capabilities. On HTTP, the response status **MUST** be `400 Bad Request`." The conditional at the front is what matters here, though: the obligation only arises for a server that has some request whose processing needs a client capability.

This server has none. It requires no client capability for anything it implements, so `-32021` is unreachable, and the PR does not add a code path that emits it. What it adds is `MissingRequiredClientCapabilityErrorData` in `internal/mcp/types.go`, a payload type carrying its own comment saying exactly that: "This server never emits this error today (see RequestMeta), but the type exists so a future request type that does need a client capability has it ready." `docs/developers/mcp-protocol.md` already records the same thing in its error-code table.

An OpenAPI document describes the responses an endpoint can actually produce. Listing `-32021` under the 400 alongside `-32020`, `-32602`, and `-32022`, all of which this server really does emit, would tell a client to expect a response it can never receive, and would be a false statement about the endpoint rather than a gap being closed. It becomes correct the day a request type here needs a declared capability, and the payload type is sitting ready for it. Not changing this; resolving.

dpage (approved):
Approving. The two findings from my review on 4 August are both properly addressed in 92ddcadf, and rather more thoroughly than I asked for.

On the HTTP status, the fix is not a special case for the modern-`initialize` rejection but a `switch` on the response error code in `handleRequestHTTP`, so the unknown-method default and the "Prompts not supported" and "Database management not supported" branches are covered too. Those last two had no coverage at all previously, because the real binary always registers both providers, and they now have direct regression tests. The assertion I flagged in `test/mcp_modern_protocol_test.go` is `StatusNotFound` accordingly.

On the stdio ping, `handlePing` is on `sendResponseFor`, so the two transports now agree about an identical request, and `TestModernProtocol_Ping_HasResultType` holds them to it. The header-based era detection that came with it was not something I raised, but it is a good catch: an HTTP request sending `MCP-Protocol-Version` whilst omitting or misspelling that rather long `_meta` key was being served as legacy.

Two things I have done to this branch directly, both flagged on the threads rather than slipped in quietly.

First, c6a896c documents the `Accept` header in the OpenAPI specification, which CodeRabbit was right to flag: the transport requires a conforming client to list both `application/json` and `text/event-stream`. The requirement binds the client rather than the server, so it is documented rather than enforced, and the parameter description says as much instead of implying a rejection this server never performs. Its counterpart suggestion, documenting `-32021` as a 400, I declined: this server requires no client capability for anything it implements, `MissingRequiredClientCapabilityErrorData` exists only as a payload type awaiting a future need, and listing an unreachable error alongside the three that genuinely occur would misdescribe the endpoint.

Second, 3b07040 merges `main`. The changelog had conflicted with entries from #221 and #222, which not only blocked merging but silently stopped CI: GitHub cannot compute a merge ref for a conflicting pull request, so nothing had run since 3a5d037b on 4 August, and the green ticks sitting on this PR were two days and two commits stale. The conflict was additive on both sides and every entry survives unaltered. Revert the merge and rebase instead if you would rather own that yourself.

All twelve checks are green on 3b07040. MCP Server (17) failed once on the way there with `TestEnsureMetadataFor_SharesFailedReload`, which is the known flake from #222 that #228 fixes, is unrelated to this work, and passed on re-run.

The `Origin` header gap recorded under "Known gaps" is a genuine DNS-rebinding exposure that the transport requires servers to close. It predates this work and is rightly out of scope here, but it wants its own issue rather than a documentation bullet that quietly ages.

PR #224, opened 2026-08-06
fix(web): label the header conversation action as Clear, not Delete

### LINKED ISSUE #223: Conversation delete does not remove the conversation
1. Open the web client and create two chat conversations.
2. Select one conversation from the left conversation tree.
3. Click the delete option from the top menu (with the save conversation option present).

**Expected result:** The selected conversation is deleted and no longer appears in the conversation list.

**Actual result:** The conversation is not deleted; it remains visible in the conversation list after the delete action.

Image

dpage:
## Summary

- The "Delete conversation" item in the status banner's conversation
  actions menu was wired to the chat interface's clear handler, which
  empties the chat window and starts a new conversation but never removes
  the conversation from the conversation list, so the reported failure was
  a mislabelled action rather than a broken delete. The dialog body
  already admitted as much ("This clears the entire current conversation
  and starts a new one") whilst the menu item, its icon, and its confirm
  button all promised a deletion.

- Deletion already exists per conversation in the conversation list, so
  this renames the header action to "Clear conversation", swaps the bin
  icon for a clear-all icon, drops the destructive red styling for the
  ordinary menu and confirm colours used elsewhere, and rewords the dialog
  to say that the conversation stays in the list where it can be reopened
  or deleted.

## Test plan

- [x] `web`: existing `StatusBanner` tests updated for the new label,
      dialog text, and confirm button, plus a new test asserting the
      header menu offers no delete action at all.
- [x] `npm test` in `web/`: 128 tests across 13 files pass.
- [x] `npm run build` and `make lint-client` clean.

Closes #223

## Summary by CodeRabbit

- **New Features**
  - Renamed the chat action from “Delete conversation” to “Clear conversation.”
  - Clearing now resets the chat and starts a new conversation while preserving the saved conversation.
  - Permanent deletion remains available through the conversation’s delete control.

- **Bug Fixes**
  - Updated confirmation messaging and controls to clearly distinguish clearing from permanent deletion.

- **Documentation**
  - Updated the changelog to describe the revised conversation actions.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -155,6 +155,15 @@ and this project adheres to
 
 ### Fixed
 
+- The conversation actions menu in the status banner header no longer
+  offers a "Delete conversation" item that does not delete anything. That
+  action resets the chat window and starts a new conversation; it never
+  removed the conversation from the conversation list, so a selected
+  conversation appeared to survive the delete. The item is now labelled
+  "Clear conversation", and its confirmation dialog explains that the
+  conversation stays in the list, where the per-conversation delete button
+  removes it for good. (#223)
+
 - Tool calls made over the streaming chat endpoint are no longer silently
   dropped, which showed up in the web client as "No response received"
   whenever a provider decided to call a tool. The terminating `done` frame
--- a/web/src/components/StatusBanner.jsx
+++ b/web/src/components/StatusBanner.jsx
@@ -42,7 +42,7 @@ import {
     Warning as WarningIcon,
     MoreVert as MoreVertIcon,
     SaveAlt as SaveIcon,
-    Delete as DeleteIcon,
+    ClearAll as ClearIcon,
 } from '@mui/icons-material';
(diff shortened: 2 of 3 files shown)


dpage:
On CodeRabbit's linked-issue warning: this is deliberate, and worth spelling out for whoever reviews.

The header menu action was never a delete. It is wired to the chat interface's `handleClear`, which empties the message list and calls `startNewConversation()`; nothing is ever removed from the conversation list, and the dialog body already said as much whilst the item, its icon, and its confirm button all promised a deletion. Deleting a conversation is already possible, per conversation, from the bin icon on each item in the conversation list, and there is a "Delete All Conversations" action beneath it, so adding a second delete path in the header would duplicate that rather than fill a gap.

The reproduction in #223 is therefore the label lying about what the action does, not a delete that fails, so the fix is to describe it accurately.

moizpgedge (approved):

PR #229, opened 2026-08-06
fix(web): derive the client version from package.json

dpage:
## Summary

`CLIENT_VERSION` was a literal in `web/src/lib/mcp-client.js`, under the comment
"keep in sync with server releases", and it had sat at `1.0.0-alpha5` ever since
that release: alpha6, beta1, beta2, beta3 and 1.0.0 all shipped without it
moving, whilst `ServerVersion`, the CLI's `ClientVersion` and `web/package.json`
were each carried along. Hence the help panel reading `Web Client:
v1.0.0-alpha5  |  Server: v1.0.0`.

It is more than cosmetic: `mcp-client.js:106` sends the same constant as
`clientInfo.version` in every MCP `initialize` handshake, so the server's logs
and traces have been recording this client as five releases older than it is.

It is now injected from `web/package.json` by a `define` in `vite.config.js`,
mirrored in `vitest.config.js` so the tests exercise the same value. A release
that bumps the manifest, which it must do anyway, now bumps this with it.

## Also: a documented list of version locations

`.claude/CLAUDE.md` gains a Versioning section recording the minimal set of
places a release must update, which I established by auditing the tree rather
than by guessing: `web/package.json` (+ lockfile), `ServerVersion`,
`ClientVersion`, `docs/changelog.md`, the Helm chart's `version` and
`appVersion`, and the git tag that the RPM and Debian builds derive from. It
also records what deliberately needs no update: this newly-derived
`CLIENT_VERSION`, and the Dockerfile `version` labels, which carry only
`major.minor`.

Per your request it states plainly that the list is a starting point rather than
authoritative, and that a manual `git grep` for the outgoing version is still
required. A list like that going stale is the same failure mode as the constant
it describes, so it should not be trusted on its own.

## Test plan

- [x] Two new tests: `CLIENT_VERSION` equals `package.json`'s version, and is a
      non-empty semver-shaped string. The first fails if a literal is
      reintroduced and drifts (verified: reverting to `'1.0.0-alpha5'` fails
      with `expected '1.0.0-alpha5' to be '1.0.0'`); the second catches the
      define being dropped from a config, which would otherwise surface as
      `undefined` in the handshake rather than failing anything.
- [x] `npm test`: 129 tests pass. `npm run build` clean.
- [x] Inspected the built bundle: no `1.0.0-alpha5` anywhere, and the manifest
      version is inlined (`mh="1.0.0"`).
- [x] Verified in a browser against a running server: the help panel now reads
      `Web Client: v1.0.0  |  Server: v1.0.0`.

## Note on the Go side

The CLI and server have the same class of hazard, and I have deliberately left
it out of this PR: see the comment below, since it needs a decision rather than
just a patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronized the web client’s displayed version and MCP handshake version with the application release version.
  * Ensured version information is generated consistently for production builds and tests.

* **Tests**
  * Added validation confirming the client version matches the application version and follows semantic versioning.

* **Documentation**
  * Documented the release versioning process and updated the changelog with the version synchronization fix.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -207,6 +207,56 @@ At the end of each README:
 
 - Update `changelog.md` with notable changes since the last release.
 
+## Versioning
+
+A release must update the version in every place below. **Treat this list
+as a starting point and not as authoritative**; it records where versions
+lived when it was written, and a new one can appear at any time without
+this file being updated. Always finish with a manual check, searching the
+tree for the outgoing version string and reading what each hit means:
+
+```bash
+git grep -n "1\.2\.3"   # the version being replaced
+```
+
+The minimal set of places to update:
+
+- `web/package.json`, along with the `version` fields in
+  `web/package-lock.json`. Run `npm version <version> --no-git-tag-version`
+  in `web/` rather than editing them by hand.
+
+- `internal/mcp/server.go`, in the `ServerVersion` constant. The server
+  reports it over MCP `initialize`, on `/health`, and in the OpenAPI
+  specification.
+
+- `internal/chat/mcp_client.go`, in the `ClientVersion` constant. The CLI
+  reports it in `--version` and its welcome banner.
+
+- `docs/changelog.md`, which needs a new release heading with the date, and
+  the entries moved out of `[Unreleased]`.
+
+- `examples/helm/pgedge-nla/Chart.yaml`, in both `version` and
+  `appVersion`.
+
+- The git tag itself, which is what the release workflow and the RPM and
+  Debian packaging derive their version from: `pkg/build-rpm.sh` and
+  `pkg/build-deb.sh` read it out of the tag, so neither needs editing.
+  Note that `pkg/common.sh` carries a hardcoded fallback for when
+  `COMPONENT_VERSION` is unset, which affects local package builds only;
+  update it if you rely on those.
+
+Two things deliberately need no update, and should not be given one:
+
+- The web client's `CLIENT_VERSION` in `web/src/lib/mcp-client.js` is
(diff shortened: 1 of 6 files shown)


dpage:
On the Go side, which I left out of this PR because it needs a decision rather than a patch.

Both `ServerVersion` and the CLI's `ClientVersion` are hand-maintained literals with exactly the hazard this PR removes from the web client; they happen to be correct today, but nothing ties them to a release either. More to the point, the release build already tries to inject the real version and fails silently: both goreleaser configs pass

```
-X main.version={{.Version}}
-X main.commit={{.Commit}}
-X main.date={{.Date}}
-X main.builtBy=goreleaser
```

for both binaries, and neither `cmd/pgedge-pg-mcp-svr/main.go` nor `cmd/pgedge-pg-mcp-cli/main.go` declares any of those variables. The Go linker ignores `-X` against a symbol that does not exist, so all four are inert: `pgedge-nla-cli --version` prints the hardcoded `chat.ClientVersion` in released builds, and the tag, commit and build date never reach either binary.

The fix is to declare those variables in both `main.go` files and have the two constants default to them, at which point the tag becomes the single source of truth for the Go side and the CLAUDE.md list in this PR loses two more entries. The decision it needs is what a plain `go build` with no ldflags should report: `dev` is the honest answer and removes the last hand-maintained literal, but it changes what `/health` and `--version` say for local builds, and a couple of tests assert the current value. Injecting `git describe` from the Makefile for local builds would give something more useful than `dev`.

Happy to do it as its own PR, though it touches the release pipeline, so it may be one for after the current release rather than during it.

moizpgedge (approved):

Issue #235, opened 2026-08-06
Gemini provider:  tool calls with errors enter infinite retry loop


imtiazqa:
When using the Gemini LLM provider, tools that return an error
(isError: true) enter an infinite retry loop instead of returning a
response to the user. The request must be manually cancelled. The same
tools work correctly with the Anthropic provider.

## Configuration

```
llm:
  enabled: true
  provider: gemini
  model: gemini-2.5-flash
  gemini_api_key_file: "/tmp/.gemini-api-key"
  per_attempt_timeout: 60
```

## Tools Tested

| Tool                      | Result        |
|---------------------------|---------------|
| list_database_connections | Pass          |
| get_schema_info           | Pass          |
| read_resource             | Infinite loop |
| count_rows                | Infinite loop |
| query_database            | Infinite loop |

## Steps to Reproduce

1. Configure the server with provider: gemini, model: gemini-2.5-flash
2. Open the web client and authenticate
3. Ask: "Read the system info resource"
   → read_resource loops indefinitely
4. Ask: "How many rows are in the docs table?"
   → count_rows loops indefinitely
5. Ask: "Query the primary database: SELECT version()"
   → query_database loops indefinitely

## Actual Result

Tools enter an infinite retry loop (40+ calls observed) until the
user manually cancels. Example:

  read_resource ❌ → read_resource ❌ → read_resource ❌ → ...

## Expected Result

All tools return a result and the LLM responds to the user .

## Possible Root Cause

The pgedge-go-llm-lib proxy converts MCP tool results into Gemini's
functionResponse format. When a tool returns isError: true, the proxy
likely passes it without Gemini-specific error formatting. Gemini does
not recognize isError: true as a terminal failure and treats it as an
unresolved function call to retry. Anthropic natively understands
isError: true as part of the MCP spec.

Gemini expects errors formatted as:

  {
    "functionResponse": {
      "name": "tool_name",
      "response": {
        "error": "error message here"
      }
    }
  }

dpage:
I have dug into this and there are two separate findings, so I want to record both rather than quietly closing it.

[...]

PR #240, opened 2026-08-10
fix(web): stop repeating a tool call that keeps failing

dpage:
## Summary

Issue #235 reports that with the Gemini provider a failing tool is called over
and over until the user cancels. It is not quite infinite: both agentic loops
are bounded by `MAX_AGENTIC_LOOPS` (50), so what actually happens is fifty round
trips, each costing a request, ending in a bare `Maximum tool execution loops
reached` error rather than anything the user can act on.

- Each executed call is keyed by tool name plus its arguments, serialised so key
  order is irrelevant, and three consecutive failures of that key end the turn
  with an ordinary assistant message naming the tool and quoting the error.
- A retry with different arguments is a different key and is left alone, since a
  model adapting its approach is making progress. A success clears the count for
  that call, so intermittent failures never accumulate.
- Nothing is thrown, so the user reads an explanation rather than an error
  banner. The guard lives in `web/src/utils/toolRepeatGuard.js` and is wired into
  both agentic loops.

## On the root cause, and why this says Refs rather than Closes

This is deliberately a backstop rather than a diagnosis, so I have not marked the
issue auto-closing.

The provider side is addressed separately in pgEdge/pgedge-go-llm-lib#34: Gemini
was receiving failed tool results as though they had succeeded, because the
provider never read `IsError`. That is a real bug with a demonstrable
consequence, verified against the live API, where a failed tool with no error
text made the model reply "I was able to read the system info resource, but it
appears to be empty" instead of admitting the failure.

However, it does not by itself reproduce the runaway. Driving the real Gemini API
through the proxy exactly as this client does, including the `sseChat` `tool_use`
inference, the model stops after a single failing call. I tried four variants
(direct client, faithful proxy round-trip, empty error text, and a mismatched
`tool_use_id`) and none looped. Something in the reporter's environment differs,
most likely the full tool set, the real system prompt, or the specific error text
a given tool emits, and a trace of the request bodies from a looping session
would settle it.

Bounding the damage is worth having regardless, and it protects against any
provider behaving this way rather than only the one reported.

## Test plan

- [x] 23 unit tests for the guard: key stability under argument reordering,
      per-call counters, reset on success, and cycle-safe serialisation.
- [x] 6 integration tests driving the real component loop with `sseChat` and
      `mcpClient.callTool` stubbed, asserting the tool is never invoked a fourth
      time, that differing arguments do not trip the guard, that fail/fail/
      success/fail/fail completes normally, that a clean conversation is
      unaffected, and that the prompt-driven loop is guarded identically.
- [x] `make lint` (0 issues) and `make test` fully green, 210 web tests passing.

Refs #235

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented repeated retries of identical failed tool calls after three consecutive failures.
  - Included user-declined executions and displayed the affected tool and error.
  - Allowed calls with different arguments to continue and reset failure counts after success.
  - Applied the safeguard to chat and prompt-driven interactions.

- **Documentation**
  - Added a changelog entry describing the repeated-failure safeguard.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -76,6 +76,16 @@ and this project adheres to
 
 ### Fixed
 
+- The web client no longer keeps re-running a tool call that cannot
+  succeed. Where a model responded to a tool error by reissuing the
+  identical call, the agentic loop would repeat it up to fifty times,
+  spending a request on each attempt, before failing with a bare
+  message about the loop limit. The loop now stops after the same tool
+  fails three times in a row with the same arguments, and explains what
+  happened, naming the tool and quoting the error. A retry with
+  different arguments is not affected, and a success clears the count,
+  so a model that adapts is left alone.
+
 - Switching `embedding.provider` (or `knowledgebase.embedding_provider`)
   away from Ollama without also setting a model no longer sends the
   Ollama model name to the new provider. `defaultConfig` seeded both
--- a/web/src/components/ChatInterface.jsx
+++ b/web/src/components/ChatInterface.jsx
@@ -974,6 +974,14 @@ const ChatInterface = ({ conversations }) => {
                     // once per block.
                     const toolResultMessages = [];
                     for (const toolUseBlock of toolUses) {
+                        // A batch of parallel tool_use blocks can repeat the
+                        // same failing call several times in one response;
+                        // stop issuing further calls the moment the guard
+                        // has already tripped on an earlier one this batch.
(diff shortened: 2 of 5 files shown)


dpage:
@coderabbitai review

dpage:
@coderabbitai review

moizpgedge (approved):

Issue #238, opened 2026-08-07
add Gemini LLM provider support to the CLI client


imtiazqa:
The MCP server supports Gemini as an LLM provider (provider: gemini in YAML config, PGEDGE_GEMINI_API_KEY env var), but the CLI client (pgedge-pg-mcp-cli) does not expose it. The --llm-provider flag only accepts anthropic, openai, or ollama. There is no --gemini-api-key flag and /set provider gemini is not available in the interactive CLI.

**Expected fix:**

Add --gemini-api-key / --gemini-api-key-file CLI flags
Add gemini to --llm-provider accepted values
Add gemini to /set provider in the interactive session
Update /help text to include gemini in the provider list (or make it dynamic)

imtiazqa:
./bin/pgedge-nla-cli   --llm-provider gemini   --gemini-api-key-file /tmp/.gemini-api-key   --mcp-mode http   --mcp-url http://localhost:8080   --mcp-auth-mode user   --mcp-username admin   --mcp-password secret123

          _
   ______/ \-.   _           pgEdge Natural Language Agent
.-/     (    o\_//           CLI: v1.0.0  Server: v1.0.0
 |  ___  \_/\---'            Type /quit to leave, /help for commands
 |_||  |_||

System: Connected to pgedge-postgres-mcp (13 tools, 1 resources, 4 prompts)
System: Using LLM: gemini (gemini-2.5-flash)
System: Database: primary (read-only)
────────────────────────────────────────────────────────────────────────────────
You: What PostgreSQL version am I running?
 → Executing tool: read_resource (pg://system_info)
                                        

Assistant:
  You are running PostgreSQL 18.4.                                                                                    
────────────────────────────────────────────────────────────────────────────────
You: /set provider gemini
System: LLM provider set to: gemini (model: gemini-2.5-flash)
You: 
You: /quit


You: /help

Available Commands:

  Navigation:
    /help                       Show this help message
    /clear                      Clear the screen
    /quit, /exit                Exit the CLI

  LLM Settings:
    /list providers             List available LLM providers
    /list models                List available models for current provider
    /set provider <name>        Set LLM provider (anthropic, openai,
                                gemini, ollama)
    /set model <name>           Set LLM model
    /show provider              Show current LLM provider
    /show model                 Show current LLM model

  Database:
    /list databases             List available database connections
    /set database <name>        Select a database connection
    /show database              Show current database connection

  MCP Resources:
[...]

PR #243, opened 2026-08-10
chore: even up the provider credential options

dpage:
> **Stacked on #241.** The base branch is `feat/issue-238-cli-gemini`, because this
> builds on the `chat.ReadAPIKeyFile` export and the Gemini config fields added
> there. Merge #241 first; GitHub will retarget this to `main` automatically.

## Summary

An audit of how API keys can be supplied turned up two asymmetries. Both fixes
are purely additive and change no existing behaviour.

**Gemini had no way to reach a proxy.** Every other LLM provider carries a base
URL override on both the server and the CLI, but neither `LLMConfig` had a
`GeminiBaseURL` field and no environment variable set one. The server's
`providers["gemini"]` block did not pass a `BaseURL` either, whilst the
anthropic, openai and ollama blocks all did, so the wiring was missing as well
as the setting. Both sides now read `gemini_base_url`, or
`PGEDGE_GEMINI_BASE_URL` from the environment.

**The CLI offered a key-file flag for Gemini alone**, even though the config
file has honoured `anthropic_api_key_file` and `openai_api_key_file` all along.
All three providers now have both flags, behaving identically: the file named on
the command line is read where the flag is applied (the loader resolves
configured key files before any flag is seen), a direct key beats one read from
a file, and a command-line path that cannot be read or yields an empty key is an
error rather than the silent fallback the loader allows for a configured path.
That logic moves into `applyAPIKeyFlags`, which returns an error and leaves
`os.Exit` to `main`, making it testable without contorting the flag parsing.

## Note on the naming, and on #239

`PGEDGE_GEMINI_BASE_URL` was deliberately left unused when the embedding side
took the `PGEDGE_GEMINI_EMBEDDING_BASE_URL` form, so this claims a reserved name
rather than renaming anything.

Worth flagging for merge order: the Gemini *embedding* support lives in #239,
which is still open, so `EmbeddingConfig.GeminiBaseURL` does not exist on this
branch at all. The two are complementary by design, but both touch
`internal/config/config.go` and the same documentation tables, so expect textual
conflicts depending on merge order. The tests here assert the two namespaces stay
independent in the only direction currently observable, with comments saying so,
so whoever merges second can tighten the other half.

Whilst updating the configuration tables it also became apparent that the
server's LLM section had never documented Gemini at all despite supporting it,
so the `gemini_api_key` and `gemini_api_key_file` rows are added alongside the
new `gemini_base_url`.

## Test plan

- [x] New config tests: `PGEDGE_GEMINI_BASE_URL` reaching both loaders, the
      server's `mergeConfig` carrying the field, and the LLM and embedding
      namespaces staying independent.
- [x] New `TestApplyAPIKeyFlags` covering all seven cases (neither flag, direct
      key, key file, both together, missing file, unreadable file, empty file)
      plus provider naming in the error text.
- [x] Flag wiring verified manually against the built binary: `-help` lists all
      six flags, the three error paths exit 1 with the right provider named, and
      a good key file gets the client past `Validate()` with no key in the
      environment.
- [x] `make lint` reports 0 issues and `make test` is fully green.

### DIFF
--- a/cmd/pgedge-pg-mcp-cli/main.go
+++ b/cmd/pgedge-pg-mcp-cli/main.go
@@ -36,7 +36,9 @@ func main() {
 	llmProvider := flag.String("llm-provider", "", "LLM provider: anthropic, openai, gemini, or ollama (default: anthropic)")
 	llmModel := flag.String("llm-model", "", "LLM model to use")
 	anthropicAPIKey := flag.String("anthropic-api-key", "", "API key for Anthropic")
+	anthropicAPIKeyFile := flag.String("anthropic-api-key-file", "", "Path to a file containing the Anthropic API key")
 	openaiAPIKey := flag.String("openai-api-key", "", "API key for OpenAI")
+	openaiAPIKeyFile := flag.String("openai-api-key-file", "", "Path to a file containing the OpenAI API key")
 	geminiAPIKey := flag.String("gemini-api-key", "", "API key for Google Gemini")
 	geminiAPIKeyFile := flag.String("gemini-api-key-file", "", "Path to a file containing the Google Gemini API key")
 	ollamaURL := flag.String("ollama-url", "", "Ollama server URL (default: http://localhost:11434)")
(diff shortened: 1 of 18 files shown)


dpage:
⚠️ **This PR currently has no CI coverage.** Worth stating plainly, because the checks list looks reassuring and is not.

Every workflow filters on `pull_request: branches: [main, develop, 'feat/llm-lib-*']`. This PR targets `feat/issue-238-cli-gemini`, so none of them triggered: no server matrix, no CLI client, no web client, no docs build. CodeRabbit likewise reports "Review skipped: reviews are disabled for this base branch". The only check that genuinely ran is Codacy.

What has actually been verified is a local run of the full gate in the worktree: `gofmt`, `go vet`, `go build`, `make lint` (0 issues) and `make test` (all Go packages plus the 181 web client tests), together with manual exercise of the six key flags against the built binary.

Once #241 merges, GitHub will retarget this to `main`, at which point CI and CodeRabbit will both run. **Please do not merge this before that has happened and gone green.** If you would rather have CI on it now, the alternative is folding these commits into #241, at the cost of mixing the tidy-up in with the issue #238 work.

PR #257, opened 2026-08-12
fix(chat): stop the model fallback choosing a non-chat model

### LINKED ISSUE #255: fix(chat): model fallback can select an embedding-only model for chat
## Summary

`selectModel` (`internal/chat/client.go`) falls back to
`availableModels[0]` with no capability check when a saved/default
model isn't available. If an embedding-only model sorts first, every
chat message fails.

## Reproduction

Ollama with only `nomic-embed-text` (embedding) and `llama3.2` (chat)
installed, saved preference pointing at a model no longer present:

```
Error: LLM error: ollama (400): "nomic-embed-text:latest" does not support chat
```

`llama3.2` was available the whole time, just not first in the list.

dpage:
## Summary

- `selectModel` fell back to the head of the provider's model list with no
  check on what kind of model it was, so an embedding model sorting first
  (Ollama's `nomic-embed-text` ahead of `llama3.2`) left every chat message
  failing whilst a usable model sat further down the same list.

- None of the four providers expose a capability flag separating chat models
  from the rest, so the fallback now skips the kinds that identify themselves
  by name: embedding, reranking, moderation, transcription, speech, image,
  video and music models. The matching is deliberately conservative, treating
  a name that carries no such marker as a chat model, because ruling out a
  usable model over a coincidence in its name would be the worse failure.

- When nothing in the list looks conversational the head is used as before,
  so the provider's own message still reaches the user rather than being
  replaced by a default it never offered. A model named on the command line,
  and a saved preference the provider still serves, are unaffected.

## Test plan

- [x] `TestIsChatCapableModel` and `TestFirstChatCapableModel` cover the name
      markers and the list walk across all four providers' naming styles.
- [x] `TestSelectModelFallbackSkipsNonChatModels` reproduces the reported
      case, and covers the provider default winning when it is served and the
      list holding nothing usable.
- [x] `make build`, `make lint` and `make test` all pass (Go suites plus 235
      web tests).

Closes #255

## Summary by CodeRabbit

- **New Features**
  - Improved automatic model selection to prioritize conversational models.
  - Skips embedding, reranking, moderation, transcription, speech, image, video, and music models during fallback selection.
  - Warns when manually selecting a model that may not support chat.

- **Bug Fixes**
  - Saved preferences for unavailable or non-conversational models now fall back to a suitable conversational model.
  - Preserves fallback behavior when no conversational models are installed.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -118,6 +118,19 @@ and this project adheres to
 
 ### Fixed
 
+- The CLI client no longer falls back to a model that cannot hold a
+  conversation. When a saved model preference named something the provider
+  no longer serves, and the provider's default was absent too, the client
+  took the first entry in the model list without checking what kind of
+  model it was; an embedding model sorting first left every message failing
+  with the provider's complaint that the model does not support chat,
+  whilst a usable model sat further down the same list. The fallback now
+  skips the model kinds that identify themselves by name, covering
+  embedding, reranking, moderation, transcription, speech, image, video and
+  music models across all four providers, and takes the first entry that
+  looks conversational. A model named on the command line, or held as a
+  saved preference the provider still serves, is used as before. (#255)
+
 - A configuration file that sets an LLM or embedding option without also
   naming the provider, or enabling the section, is no longer discarded.
   The merge tested only those two fields, so a file that configured just a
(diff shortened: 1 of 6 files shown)


dpage:
@coderabbitai review

dpage:
@coderabbitai review

moizpgedge (approved):

PR #267, opened 2026-08-13
fix(docker): create multi-database config file with mode 600

### LINKED ISSUE #261: Simple deployment mode writes the database password to a plain-text, overly-readable file
**Priority:** Low

One deployment mode (multi-database setup via environment variables, not the mode our main hosted deployments use) writes a config file that contains a real database password in plain text, and that file ends up more widely readable than it needs to be.

This does not affect our main hosted deployments, which deliver the password a safer way. It only matters for anyone using the simpler, environment-variable-driven deployment mode.

dpage:
## Summary
- The Docker entrypoint's multi-database deployment mode (`init-server.sh`) writes a generated `postgres-mcp.yaml` containing plaintext database passwords, using the process's default umask (typically leaving it `644`, group- and world-readable).
- The file is now created with mode `600` before any content is written, so the password is never briefly exposed under a wider mode.
- Single-database mode, which passes the password as a command-line argument rather than writing a file, is unaffected.

## Test plan
- [x] Extracted `generate_multi_db_config` and ran it standalone under `umask 022`, confirming the resulting file is `600` (previously `644`) and contains the plaintext password as expected.
- [x] `bash -n docker/init-server.sh` passes.

Closes #261

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security for Docker multi-database deployments by restricting generated configuration files containing database credentials to the owner.
  * Prevented credentials from being temporarily exposed through overly permissive file permissions.

* **Documentation**
  * Added a changelog entry describing the security improvement.


moizpgedge (approved):
Verified beyond the PR's own test plan:

- Confirmed single-database mode never calls `generate_multi_db_config` at all (traced the `USE_CONFIG_FILE` branch), so it's genuinely unaffected as claimed.
- Extracted the function and ran it directly under four umasks (`000`, `002`, `022`, `077`) — resulting file is `-rw-------` (600) in every case, including `000` (which would otherwise yield `666`), confirming the fix isn't umask-dependent.
- Tested the upgrade path: simulated a pre-existing `644` config left by an older, unpatched container run on a persistent volume, then ran the fixed function against it — it self-heals to `600`.
- Confirmed content correctness (real password, correct YAML) is unaffected by the permission change.
- `bash -n` and `shellcheck` clean; shellcheck's only hits are pre-existing style nits elsewhere in the file, unrelated to this change.

Approving.

moizpgedge (approved):

PR #282, opened 2026-08-25
fix: negotiate legacy MCP protocol revisions up to 2025-11-25

### LINKED ISSUE #277: pgedge-postgres-mcp server reports MCP protocol version 2024-11-05, incompatible with Streamable HTTP-only clients (e.g., OpenAI Codex)
The MCP server's initialize response always returns "protocolVersion": "2024-11-05", regardless of what protocol version the client requests. Confirmed via direct initialize calls:

- Cloud endpoint (server v1.0.0): reports 2024-11-05
- Upgraded staging test build (server v1.1.0-beta1): also reports 2024-11-05

So this isn't fixed by the version upgrade already in progress, it's present in every build tested so far.

Why it matters: "Streamable HTTP" as an MCP transport was introduced in protocol revision 2025-03-26. A server capped at 2024-11-05 predates that transport spec entirely. Clients with a strict Streamable HTTP implementation and no fallback (OpenAI Codex, confirmed via direct testing) fail to connect, silently, as a generic "server not available" message with no error surfaced to the user. Clients with a graceful fallback to the older SSE transport (Cursor, confirmed via its own connection logs showing "falling back to SSE") still work.

**Reproduction:**

curl -s -X POST <mcp-endpoint>/mcp/v1 \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"test","version":"1.0.0"}}}'
Response always shows "protocolVersion":"2024-11-05" regardless of the version requested.

**Expected:** The server should support a current MCP protocol revision, ideally matching whatever version the client requests, and support Streamable HTTP (introduced in the 2025-03-26 spec) so clients like Codex that require it can connect.

**Actual:** The server always responds with "protocolVersion": "2024-11-05", no matter what version the client asks for (tested requesting 2025-06-18, got 2024-11-05 back). True on both the live cloud endpoint (server v1.0.0) and the upgraded staging build (v1.1.0-beta1), so the in-progress version upgrade doesn't fix it.

dpage:
## Summary
- The server always answered `initialize` with `protocolVersion: "2024-11-05"`, regardless of what the client requested, because `supportedProtocolVersions` only ever had that one entry.
- Revision `2025-03-26` introduced the Streamable HTTP transport; a client with a strict implementation and no fallback (OpenAI Codex, per the issue) refused to connect outright because the reported revision predated that transport entirely.
- Extends `supportedProtocolVersions` with `2025-03-26`, `2025-06-18`, `2025-11-25` and bumps the server's default (`ProtocolVersion`) to the newest of them. `NegotiateProtocolVersion` already picks the nearest supported revision at or below the request, so this alone is enough to answer appropriately instead of always falling back to the oldest.
- This interacts with the modern (2026-07-28) stateless protocol support added since #214/#266: `isModernHTTP` treated *any* `MCP-Protocol-Version` header as a modern signal, on the documented assumption that no pre-2025-06-18 client ever sends it. That assumption breaks once the server also speaks 2025-06-18/2025-11-25, which require that same header (with a legacy dated value) on every request. `isModernHTTP` now checks the header's *value* against the supported legacy revisions rather than just its presence, so a real legacy Streamable HTTP client isn't misclassified as modern and rejected for missing the modern-only `Mcp-Method`/`Mcp-Name` headers.

## Test plan
- [x] Extended `TestNegotiateProtocolVersion` with exact-match cases for `2025-03-26`/`2025-06-18` and a between-two-supported-revisions case.
- [x] Added `TestIsModernHTTP_LegacyProtocolVersionHeaderIsNotModern` and `TestIsModernHTTP_ModernProtocolVersionHeaderIsModern` covering the header-value disambiguation.
- [x] Confirmed every existing test that hardcodes `protocolVersion: "2024-11-05"` requests that exact (oldest) revision and still gets it back unchanged.
- [x] `go test ./internal/mcp/...` passes; `go vet ./...` clean; `go vet ./test/...` (integration tests, which need a live DB to run) compiles clean.

Closes #277

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MCP protocol negotiation across legacy and modern clients.
  - Preserved compatibility with supported protocol revisions, including `2025-03-26`, `2025-06-18`, and `2025-11-25`.
  - Fixed initialization handshakes and legacy requests being incorrectly treated as modern HTTP requests.
  - Improved compatibility with Streamable HTTP transport and future protocol versions.
  - Enforced database query row limits and improved `SELECT` and `FETCH` limit handling.
  - Corrected knowledgebase default configuration behavior.

- **Documentation**
  - Added changelog information covering these fixes and compatibility improvements.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -38,6 +38,13 @@ and this project adheres to
   variant; PostgreSQL rejected the two conflicting clauses outright
   either way.
 
+- The server now negotiates the legacy MCP handshake up to revision
+  `2025-11-25`, instead of always answering `2024-11-05` regardless of
+  what a client requests. This includes revision `2025-03-26`, which
+  introduced the Streamable HTTP transport: strict clients with no
+  fallback (e.g. OpenAI Codex) previously refused to connect because
+  the reported revision predated that transport outright.
+
 ## [1.1.0] - 2026-08-17
 
 ### Added
--- a/internal/mcp/http_server.go
+++ b/internal/mcp/http_server.go
@@ -331,12 +331,14 @@ func (s *Server) handleHTTPRequest(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// A request carrying io.modelcontextprotocol/protocolVersion in its
-	// _meta -- or, on this transport, just an MCP-Protocol-Version
-	// header, which no pre-2025-06-18 client ever sends -- is modern
-	// (2026-07-28, stateless per-request negotiation); a request with
-	// neither, including every initialize handshake, is legacy and
-	// reaches handleRequestHTTP exactly as before this server added
-	// modern support. See isModernHTTP in modern.go.
+	// _meta -- or, on this transport, an MCP-Protocol-Version header
+	// whose value isn't one of this server's supported legacy revisions
+	// -- is modern (2026-07-28, stateless per-request negotiation); a
+	// request with neither, including every initialize handshake and
+	// every legacy Streamable HTTP request carrying its own negotiated
+	// legacy version in that same header, is legacy and reaches
+	// handleRequestHTTP exactly as before this server added modern
+	// support. See isModernHTTP in modern.go.
 	//
 	// preflightRejected distinguishes a rejection at this stage (header
 	// mismatch, missing required _meta field, unsupported version) from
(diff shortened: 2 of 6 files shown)


dpage:
@coderabbitai review

moizpgedge (approved):

PR #285, opened 2026-09-09
docs: separate Jinja2 raw markers from their code fences

### LINKED ISSUE #283: Blank line missing before {% endraw %} causes it to vanish when attr_list is enabled downstream
Three pages in `docs/` have a `{% endraw %}` marker on the line directly after a closing code fence, with no blank line separating them:

| File | Fence closes | `{% endraw %}` |
|---|---|---|
| `docs/reference/tools.md` | line 1015 | line 1016 |
| `docs/guide/cli-client.md` | line 387 | line 388 |
| `docs/contributing/ci-cd.md` | line 355 | line 356 |

(Line numbers from `main`; the same pattern is also present at `v1.0.0`, three lines earlier in each file since `tools.md` and `cli-client.md` have since grown.)

## Effect downstream, and why it's worth fixing regardless

These markers look like Jinja2's `{% raw %}`/`{% endraw %}` tags, presumably meant to protect the fenced block's contents from template processing somewhere in the docs pipeline. Plain [Python-Markdown](https://python-markdown.github.io/) doesn't process Jinja syntax, so today they render as literal, visible text: a stray `{% raw %}` paragraph before the block and a stray `{% endraw %}` paragraph after it. That's already wrong, independent of any downstream tooling.

It became a visible rendering bug on `docs.pgedge.com`, which is what surfaced this: that site recently started enabling the `attr_list` markdown extension so that content from other pgEdge repos (which rely on it) renders correctly. `pymdownx.superfences`, used for fenced code blocks, doesn't force a paragraph break after a closing fence when the next line isn't separated by a blank line, so the fence and `{% endraw %}` get parsed as one paragraph. `attr_list` then treats a trailing `{...}`-shaped chunk at the end of a paragraph as an attribute list for that paragraph, consumes `{% endraw %}`, and it vanishes entirely, replaced by bogus attributes on the wrapping tag (`<p _="%" endraw="endraw">`) rather than appearing as text.

Confirmed with a minimal reproduction:

```python
import markdown
html = markdown.markdown(
    "```bash\ncode\n```\n{% endraw %}\n",
    extensions=['pymdownx.superfences', 'attr_list'],
)
# <p _="%" endraw="endraw"><div class="highlight">...</div></p>
```

## Fix

Add a blank line between the closing fence and `{% endraw %}` in each of the three files (and correspondingly before the matching `{% raw %}`, if that isn't already blank-line-separated — it is in the cases checked). That alone is enough to stop both problems: the marker renders as its own paragraph under plain Python-Markdown, and it won't be silently eaten if `attr_list`-style extensions are ever added to whatever renders these docs.

Whether the `{% raw %}`/`{% endraw %}` markers should be there at all, if nothing downstream currently consumes them as Jinja, is a separate question — happy to raise that separately if useful, but the blank-line fix is safe either way.

dpage:
## Summary

- Adds a blank line between the closing code fence and the Jinja2 raw
  end marker in `docs/reference/tools.md`,
  `docs/guide/cli-client.md` and `docs/contributing/ci-cd.md`, and one
  after the opening marker in the two files that lacked it.
- Without the separation the fence and the marker parse as a single
  paragraph, so a renderer with `attr_list` enabled reads the marker as
  an attribute list and swallows it, emitting `<p _="%"
  endraw="endraw">` instead of the marker's text; that is what surfaced
  on docs.pgedge.com.
- The markers themselves stay: the wrapped blocks contain shell
  here-document redirections and pgvector's `<#>` operator, which
  collide with the `<<` and `<#` delimiters this repo configures
  `mkdocs-macros` with, so the raw blocks are load-bearing here.

## Test plan

- [x] `mkdocs build` succeeds and the built pages contain no literal
      raw markers, matching the behaviour before the change.
- [x] Rendering each page through `markdown` with
      `pymdownx.superfences` and `attr_list` no longer produces the
      bogus `endraw="endraw"` attribute.

Closes #283

## Summary by CodeRabbit

* **Documentation**
  * Improved formatting of templated code examples across the documentation.
  * Prevented raw-block markers from being misinterpreted as code-fence attributes.
  * Updated the unreleased changelog with these documentation corrections.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,16 @@ and this project adheres to
 
 ### Fixed
 
+- The Jinja2 raw block markers that shield three code blocks from the
+  documentation site's macro processing are now separated from their
+  fences by blank lines. Without the separation the closing fence and
+  the marker after it parse as a single paragraph, so a renderer with
+  the `attr_list` markdown extension enabled treats the marker as an
+  attribute list for that paragraph and swallows it, emitting bogus
+  attributes on the wrapping tag instead. The affected pages are
+  `docs/reference/tools.md`, `docs/guide/cli-client.md` and
+  `docs/contributing/ci-cd.md`.
+
 - The release workflow now names its archives after the tag that
   triggered it. GoReleaser was left to work the version out for itself,
   which it does by asking git which tags point at the commit and taking
--- a/docs/contributing/ci-cd.md
+++ b/docs/contributing/ci-cd.md
@@ -329,6 +329,7 @@ jobs:
 ### Setup
 
 {% raw %}
+
 ```bash
 # Install pre-commit hook
 cat > .git/hooks/pre-commit << 'EOF'
@@ -353,6 +354,7 @@ EOF
 
 chmod +x .git/hooks/pre-commit
 ```
+
 {% endraw %}
 
 ### Skip Hook (When Needed)
--- a/docs/guide/cli-client.md
+++ b/docs/guide/cli-client.md
@@ -368,6 +368,7 @@ ollama pull llama3
 Create a configuration file and use it:
 
 {% raw %}
+
 ```bash
 # Create config file
 cat > .pgedge-pg-mcp-cli.yaml << EOF
@@ -385,6 +386,7 @@ EOF
 # Run with config file
 ./bin/pgedge-nla-cli -config .pgedge-pg-mcp-cli.yaml
 ```
+
 {% endraw %}
 
 ## Interactive Commands
--- a/docs/reference/tools.md
+++ b/docs/reference/tools.md
@@ -1013,5 +1013,6 @@ approximate nearest neighbor search...
 
 Total: 5 chunks, ~687 tokens
 ```
+
 {% endraw %}
 


AntTheLimey (approved):
This looks right. The mkdocs.yml macro delimiters (`<<`/`>>` for
variables, `<#`/`#>` for comments) confirm the raw markers are
necessary here: `tools.md` wraps pgvector's `<#>` operator and the
other two pages wrap `<< 'EOF'` heredocs, all of which would otherwise
collide with the custom delimiters. `mkdocs build` succeeds on both
main and this branch, the built HTML for all three affected pages is
clean of leaked markers on the branch, and a full-tree grep turns up
no other page with the same marker-adjacent-to-fence pattern, so the
fix is complete. The changelog entry is accurate and well within the
79-character wrap. One thing I could not reproduce: the bogus
`endraw="endraw"` attribute does not appear in a standalone build of
this repo's own mkdocs.yml, with or without `attr_list` added, so the
defect looks specific to the docs.pgedge.com aggregate build rather
than anything this checkout renders. That doesn't change my read of
the fix — the separation is correct either way — but it does mean the
regression is only confirmed fixed once the published site rebuilds.
Neither sibling PR (#286, #288) touches the raw-block region of
cli-client.md, so merge order is unconstrained.

